PR #33, opened 2011-04-11
Frozen uri

gkellogg:
Recent changes to Addressable::URI using the @var ||= value pattern. This causes problems when the URI is frozen and such accessors are used, especially when method is @var ||= nil. These changes address this by either returning @var if it is defined, or avoiding creating the instance variable if it would just be nil.

Also, add missing UTF-8 encoding to specs.


macotouenaca (commented):
ビックマネー

PR #133, opened 2013-07-30
Addressable::Template: Substitution with something that is not a Hash

levinalex:
I'm trying to replace my ad-hoc URI-Template parser with Addressable::Template.

However, I don't have the substitutions in Hash form. The current Template implementation assumes that substitutions are Enumerable (it calls `inject` on `mapping`)

Not calling inject makes it possible to pass a lambda (or anything else that responds to `[]` as a substitute:

```
describe "Substituting from an optional block" do
  sub = lambda { |x| "foo-#{x}" }
  Addressable::Template.new("{bar}/{baz}").expand(sub).to_str.should == "foo-bar/foo-baz"
end
```

Ideally I'd like to change expand to take a block like so:

```
template.expand { |field| "foo-#{field}" }
```

but I'd like to hear what you think about changing the API in that way.

### DIFF
--- a/lib/addressable/template.rb
+++ b/lib/addressable/template.rb
@@ -829,6 +829,7 @@ module Addressable
     # @return [Hash]
     #   A hash with stringified keys
     def normalize_keys(mapping)
+      return mapping unless mapping.respond_to?(:inject)
       return mapping.inject({}) do |accu, pair|
         name, value = pair
         if Symbol === name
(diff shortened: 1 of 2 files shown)


sporkmonger:
Also interested in @therabidbanana's take on this PR.

therabidbanana:
This seems entirely reasonable to me - documentation would probably be good.

sporkmonger:
@levinalex I like this change, but I'd like to see at least one more test covering the "anything else that responds to `[]`" scenario as well as a documentation update, as @therabidbanana mentioned.

mrttpz (approved):

PR #165, opened 2014-08-06
Remove some Rubyforge references

abotalov:
Also I have some further questions:
1. As code documentation is written in YARD, do you agree that RDoc-related stuff in `yard.rake` can also be removed?
2. Do you want Addressable 3.0 to support ruby 1.8.7/ree? Or do you want to specify required_ruby_version as >= 1.9.3?
3. Probably there's no more reason to use `tasks/gem.rake` as rubyforge is dead - gemspec contents can be stored directly in `addressable.gemspec` and you can use `bundler/gem_tasks` to release.
4. Is website deployed somewhere? There are some unusual advices there (like `sudo gem install`, use submodules instead of `gem` with `:git` option). Probably website folder can be removed too.


sporkmonger:
Addressable has a lot of Rake crud that's left over from having been born circa 2006 unfortunately. The code living within `lib` is super-mature; everything outside `lib` has bit-rotted a little and needs me to give it some serious love.

I ripped the last vestiges of Rubyforge out of my Rakefile awhile ago, but yeah, the `website/` folder is left over from when addressable.rubyforge.org was still a thing. These days, hosting a project page for addressable seems pretty unnecessary. The stuff that lived on the rubyforge website has now been superceded by things like Coveralls, Gemnasium, etc for the most part.

Leaving this PR open only because you raised some good points, and this is a handy reference for things I need to fix.

sporkmonger:
I'm going to close this one as obsolete. I think everything's already taken care of.

sporkmonger:
Well, except for `gem.rake` anyway.

PR #166, opened 2014-09-03
Convert specs to RSpec 3.0 with Transpec

petergoldstein:
1. Ran transpec to convert the specs to RSpec 3.0.x syntax
2. Fixed a couple of RSpec deprecation warnings by changing `pending` to `skip`
3. Fixed an issue which was causing the build to fail.  Recent versions of Coveralls no longer support Ruby 1.8.7, so the Coveralls gem should only be included on 1.9+ platforms.

After these changes the build runs green for all required platforms, and it should be trivial to upgrade to RSpec 3.x.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 
+gemspec
+
 group :development do
   gem 'launchy'
   gem 'yard'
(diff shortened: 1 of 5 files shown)


sporkmonger:
This conversion has already happened in a different commit.

PR #167, opened 2014-09-03
Simplify gemspec

petergoldstein:
Simplify the gemspec a bit.  Eliminate duplicate gem declaration warnings.  Also includes the platforms setting that gets the build to green.


sporkmonger:
This issue has already been addressed in other commits.

PR #170, opened 2014-09-30
Fix typo in Addressable::Template extract example in README. Fixes #164

### LINKED ISSUE #164: Problem with Template#extract
The following example is from README.md and seems to be failing with 2.3.6. It does seem to work when I try 2.2.8. Am rolling back to 2.2.8 for now...

johns-mbp:test john$ ruby -v
ruby 2.1.2p95 (2014-05-08 revision 45877) [x86_64-darwin13.0]
johns-mbp:test john$ pry
[1] pry(main)> require 'addressable/uri'
=> true
[2] pry(main)> require 'addressable/template'
=> true
[3] pry(main)> template = Addressable::Template.new(
[3] pry(main)*   "http://{host}{/segments}/{?one,two,bogus}{#fragment}"  
[3] pry(main)* )  
=> #<Addressable::Template:0x3fcb77b4b950 PATTERN:http://{host}{/segments}/{?one,two,bogus}{#fragment}>
[4] pry(main)> uri = Addressable::URI.parse(
[4] pry(main)*   "http://example.com/a/b/c/?one=1&two=2#foo"  
[4] pry(main)* )  
=> #<Addressable::URI:0x3fcb77b5da4c URI:http://example.com/a/b/c/?one=1&two=2#foo>
[5] pry(main)> template.extract(uri)
=> nil
[6] pry(main)> Gem.loaded_specs['addressable'].version
=> Gem::Version.new("2.3.6")

jure:
I'm not too proud to say I've spend a good part of an hour tracking this one down. Turns out the issue that was confusing in #164 is a simple missing asterisk in the README example pattern.

P.S.: You should perhaps lock the version of RSpec to 2.11, as the >= 2.9 dependency specification in the Gemfile will just install 3.x by default, which doesn't work. But then 2.9 doesn't work either (because of `subject` and `expect` API changes).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,8 +9,9 @@ end
 
 group :test, :development do
   gem 'rake', '>= 0.7.3'
-  gem 'rspec', '>= 2.9.0'
+  gem 'rspec', '~> 2.99'
   gem 'coveralls', :require => false
+  gem 'rest-client', '~> 1.6.8'
 end
 
 gem 'idn', :platform => :mri_18
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ template.partial_expand({"one" => "1", "three" => 3}).pattern
 #=> "http://example.com/?one=1{&two}&three=3"
 
 template = Addressable::Template.new(
-  "http://{host}{/segments}/{?one,two,bogus}{#fragment}"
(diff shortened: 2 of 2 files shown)


sporkmonger:
Resolved in #183 I believe.

PR #174, opened 2014-10-29
Memoize Addressable::URI instance variables

sferik:
In the case where a fragment is present but empty (e.g. "http://example.com#"), the `normalized_fragment` method returns `nil`, so it can't be memoized with `||=`. This [patch](https://github.com/sferik/addressable/commit/67e36a4c7f406b874281037b3454d2cfd19ddc44) fixes this problem, so the `@normalized_fragment` instance variable is only ever assigned once, even when the method returns `nil`. This was causing an issue with frozen Addressable::URI (see https://github.com/sferik/t/issues/197).

I hope you’ll find a moment to release this fix during your adventures in Tanzania.

Cheers! :beers:

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -839,7 +839,8 @@ module Addressable
     #
     # @return [String] The scheme component, normalized.
     def normalized_scheme
-      self.scheme && @normalized_scheme ||= begin
+      return nil unless self.scheme
+      @normalized_scheme ||= begin
         if self.scheme =~ /^\s*ssh\+svn\s*$/i
           "svn+ssh"
         else
@@ -888,7 +889,9 @@ module Addressable
     #
     # @return [String] The user component, normalized.
     def normalized_user
-      self.user && @normalized_user ||= begin
+      return nil unless self.user
+      return @normalized_user if defined?(@normalized_user)
+      @normalized_user ||= begin
         if normalized_scheme =~ /https?/ && self.user.strip.empty? &&
             (!self.password || self.password.strip.empty?)
           nil
@@ -940,7 +943,9 @@ module Addressable
     #
     # @return [String] The password component, normalized.
     def normalized_password
-      self.password && @normalized_password ||= begin
+      return nil unless self.password
+      return @normalized_password if defined?(@normalized_password)
+      @normalized_password ||= begin
         if self.normalized_scheme =~ /https?/ && self.password.strip.empty? &&
             (!self.user || self.user.strip.empty?)
           nil
@@ -1003,7 +1008,9 @@ module Addressable
     #
     # @return [String] The userinfo component, normalized.
     def normalized_userinfo
-      self.userinfo && @normalized_userinfo ||= begin
+      return nil unless self.userinfo
+      return @normalized_userinfo if defined?(@normalized_userinfo)
+      @normalized_userinfo ||= begin
         current_user = self.normalized_user
         current_password = self.normalized_password
         if !current_user && !current_password
@@ -1058,7 +1065,8 @@ module Addressable
     #
     # @return [String] The host component, normalized.
     def normalized_host
-      self.host && @normalized_host ||= begin
+      return nil unless self.host
+      @normalized_host ||= begin
         if !self.host.strip.empty?
           result = ::Addressable::IDNA.to_ascii(
             URI.unencode_component(self.host.strip.downcase)
@@ -1156,7 +1164,8 @@ module Addressable
     #
     # @return [String] The authority component, normalized.
     def normalized_authority
-      self.authority && @normalized_authority ||= begin
+      return nil unless self.authority
+      @normalized_authority ||= begin
         authority = ""
         if self.normalized_userinfo != nil
           authority << "#{self.normalized_userinfo}@"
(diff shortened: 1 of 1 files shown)


sporkmonger:
Tests need to pass.

sferik:
@sporkmonger Working on it. It turns out that Ruby’s `defined?` keyword, passed an instance variable, behaves differently from `Object#instance_variable_defined?`. This was very surprising to me.

sferik:
@sporkmonger Never mind. That’s not actually true. I was using `defined?` incorrectly.

sferik:
Not sure what all the @coveralls noise is about but it should be good to merge now.

sferik:
Okay, now coveralls is failing but only because the total number of lines of code went down in a file with 100% c0 coverage. I don’t think that should count against this pull request. Thanks for considering it.

sporkmonger:
Wow that coveralls spam is annoying.

spec/addressable/template_spec.rb
```diff
@@ -1293,6 +1323,12 @@ def self.match(name)
     end
   end
 
+  describe 'inspect' do
+    it "returns a string representation of the MatchData's state" do
+      expect(its.inspect).to match(/\A#<Addressable::Template::MatchData:0x[0-9a-f]{0,12} RESULT:{"foo"=>"ab", "bar"=>"cd"}>\z/)
```
sporkmonger: Line is too long. Also not clear why this is related to the rest of the pull request.

sporkmonger:
I gotta tell ya, I appreciate that you put a lot of work into this, but I still can't merge this.

Drop the bundle changes, drop the gemspec and dependency changes, drop essentially everything that's not either a change to the `lib` directory or the `spec` directory, or directly dependent on a change made to one or the other.

Pull requests should always change the least amount of code possible, and they should always change only one thing at a time.

spec/addressable/template_spec.rb
```diff
@@ -728,6 +728,21 @@ def self.match(name)
...
 describe Addressable::Template do
+  describe 'initialize' do
+    context 'with a non-string' do
+      it 'raises a TypeError' do
+        expect { Addressable::Template.new(nil) }.to raise_error(TypeError)
+      end
+    end
+  end
+
+  describe 'freeze' do
```
sporkmonger: This is the only template spec that appears to be related to freezing a template. This should be the only template spec in this PR.

spec/addressable/template_spec.rb
```diff
@@ -818,6 +833,21 @@ def self.match(name)
       its(:captures){ should == ["foo", nil, nil] }
     end
   end
+
+  describe 'match' do
```
sporkmonger: These are good tests, but they should be in a different PR.

spec/addressable/template_spec.rb
sferik: This is related to the rest of the pull request insofar as Coveralls was failing because I removed a few lines of covered code, meaning the overall percentage of covered code went down. This covers some uncovered code, which making Coveralls happy again. I’ll remove it or move it into another pull request if you don’t mind merging this pull requests with Coveralls failing.

sferik:
> Pull requests should strive to change the least amount of code possible, and they should change only one thing at a time. This PR is problematic for the latter reason.

I’ve split this into three separate pull requests:

1. This pull request aims to capture only the original intent: **to memoize `Adressable::URI` instance variables**.
2. The additions to the `Addressable::Template` spec, which were only added to make Coveralls :green_heart:, have been moved into https://github.com/sporkmonger/addressable/pull/190.
3. The changes to optimize the CI build are captured in https://github.com/sporkmonger/addressable/pull/189.

I believe these three pull requests new adhere to the principle you’ve stated above but please let me know if you’d like me to subdivide these pull requests any further. Also, feel free to cherry pick individual commits from this pull request (or the others), without merging it, whole :pig2:.

Thank you for taking the time to consider these changes.

sporkmonger:
Amazing thanks so much, I know that was probably a good bit of work.

lib/addressable/uri.rb
```diff
@@ -2356,5 +2363,10 @@ def split_path(path)
       splitted << EMPTY_STR if path.end_with? SLASH
       splitted
     end
+
+    def reset_uri_string_and_hash
```
sporkmonger: Please add YARD documentation for this method.

sporkmonger:
Everything looks good except for the missing YARD documentation.

lib/addressable/uri.rb
sporkmonger: Actually, I think I'd prefer a different method name too. The only thing these two have in common is that they're essentially composite values for the entire URI. Method name should probably reflect that rather than being quite so literal. Especially since it's very possible that there would be future additions to the list and in Addressable 3, subclass overrides to this method.

`remove_composite_instance_variables` seems a little too verbose, maybe `remove_composite_values`?

sferik: I’m happy to rename the method and add YARD documentation. I’ll do that now. I was a bit fast and loose with the naming an documentation because it’s a private method.

sferik:
This pull request should be good to merge as soon as the build turns green.

sferik:
Any addition feedback on this one?

sporkmonger:
Nope! Everything looks good!

PR #186, opened 2015-03-04
Let query_values return something meaningful

tessi:
`query_values` now returns an empty version of its `return_type` instead of nil when there is no query.

So the following is `true:
`ruby
Addressable::URI.parse('http://example.com').query_values == {}
```


sporkmonger:
Wasn't an accident that it returns `nil` – that was very intentional to allow differentiation in the `query_values` return value between no query and an empty query, which are not the same thing. There are a number of real-world contexts in which you'll get the output of `query_values` sans the rest of the URI. If you return `{}` in that scenario, you effectively have information loss.

PR #188, opened 2015-03-24
Reject quotes in host name

gurdiga:


gurdiga:
Guys, I’m not sure what to do with this “Coverage decreased” failure: the [report seems to say](https://coveralls.io/builds/2182005) that coverage decreased for `lib/addressable/idna/pure.rb` and `lib/addressable/template.rb` which I didn’t touch.

Could you please guide me? (blush)

sporkmonger:
@gurdiga Updated message shows no coverage drop.

sporkmonger:
The strict definition of the `host` component is that it's either an IPv4 literal, an IPv6 literal, an IPvFuture literal, or a registered name. A registered name is defined as `*( unreserved / pct-encoded / sub-delims )` As it turns out, `sub-delims` is defined as `sub-delims = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="` and therefore the single-quote is explicitly allowed here.

gurdiga:
Adjusted to exclude single quotes.

I’m wondering in what context would those `sub-delims` be useful in a host name…? :-/

PR #189, opened 2015-04-02
Make build faster

sferik:
I’ve cherry-picked the CI build optimizations, originally included in https://github.com/sporkmonger/addressable/pull/174, into their own pull request. Please let me know if you’d like me to subdivide these any further.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,15 +2,15 @@ bundler_args: --without development --retry=3 --jobs=3
 cache: bundler
 language: ruby
 rvm:
+  - rbx-2
+  - jruby
+  - jruby-head
+  - ruby-head
   - 1.9.2
   - 1.9.3
   - 2.0.0
   - 2.1
   - 2.2
-  - jruby
-  - jruby-head
-  - rbx-2
-  - ruby-head
 matrix:
   allow_failures:
     - rvm: ruby-head
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,3 @@
-before_install:
- - sudo apt-get update
- - sudo apt-get install idn
 bundler_args: --without development --retry=3 --jobs=3
 cache: bundler
 language: ruby
@@ -19,3 +16,4 @@ matrix:
     - rvm: ruby-head
     - rvm: jruby-head
   fast_finish: true
+sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 before_install:
  - sudo apt-get update
  - sudo apt-get install idn
-bundler_args: --without development
+bundler_args: --without development --retry=3 --jobs=3
 cache: bundler
 language: ruby
 rvm:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+bundler_args: --without development
 language: ruby
 rvm:
   - 1.9.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+before_install:
+ - sudo apt-get update
+ - sudo apt-get install idn
 bundler_args: --without development
 language: ruby
 rvm:
@@ -14,6 +17,4 @@ matrix:
   allow_failures:
     - rvm: ruby-head
     - rvm: jruby-head
-before_install:
- - sudo apt-get update
- - sudo apt-get install idn
+  fast_finish: true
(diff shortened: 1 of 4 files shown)


sporkmonger:
A nice side-effect of this getting pulled out into its own PR is that it's now much clearer what you were trying to accomplish. :-)

.travis.yml
```diff
@@ -1,18 +1,19 @@
+bundler_args: --without development --retry=3 --jobs=3
+cache: bundler
 language: ruby
 rvm:
+  - rbx-2
+  - jruby
+  - jruby-head
+  - ruby-head
```
sferik: Since these builds take the longest to complete, they should be started first. The way it was working before, all the other builds would finish before the Rubinius build started and then it would take a few minutes for the entire build to complete. I wish I could draw easily insert a Gantt chart to illustrate this but I think you get the idea.

.travis.yml
```diff
@@ -1,18 +1,19 @@
...
-  - jruby-head
-  - rbx-2
-  - ruby-head
 matrix:
   allow_failures:
     - rvm: ruby-head
     - rvm: jruby-head
-before_install:
- - sudo apt-get update
- - sudo apt-get install idn
```
sferik: This `idn` package appears to be part of the default Travis Docker image, since the build is passing without it. If that ever changes and it needs to be added back, it can be done without `sudo` suing the [APT add-on](http://docs.travis-ci.com/user/apt/).

Gemfile
```diff
@@ -2,10 +2,18 @@ source 'https://rubygems.org'
 
 gemspec
 
+gem 'rake', '~> 10.4', '>= 10.4.2'
+
+group :test do
+  gem 'rspec', '~> 3.0'
+  gem 'rspec-its', '~> 1.1'
```
sferik: Moving these gems out of the `development` group in the gemspec and into their own test group allows non-test development gems to be excluded from the CI install, which speeds up the build. The only disadvantage is that `gem install addressable --development` doesn’t install all the development dependencies, but it wasn’t doing that before anyway. Working on Addressable in development requires Bundler, so I’ve made Bundle the sole development dependency, which gives us finer-grained control in the `Gemfile`. `Gem::Specification` only allows coarse-grained, development or runtime dependencies.

Gemfile
```diff
@@ -15,9 +23,4 @@ group :test, :development do
   ]
 end
 
-gem 'idn', :platform => :mri_18
```
sferik: Ruby 1.8 no longer supported.

Gemfile
```diff
@@ -15,9 +23,4 @@ group :test, :development do
...
 gem 'idn-ruby', :platform => :mri_19
-
-platforms :rbx do
-  gem 'rubysl-openssl', '2.2.1'
-end
```
sferik: This is no longer necessary. The Rubinius standard library is now packaged with the language, as it should be.

addressable.gemspec
```diff
@@ -12,32 +12,23 @@ Gem::Specification.new do |s|
...
+  s.files = ["CHANGELOG.md", "Gemfile", "LICENSE.txt", "README.md", "Rakefile", "addressable.gemspec", "benchmark", "coverage", "data", "data/unicode.data", "lib", "lib/addressable", "lib/addressable.rb", "lib/addressable/idna", "lib/addressable/idna.rb", "lib/addressable/idna/native.rb", "lib/addressable/idna/pure.rb", "lib/addressable/template.rb", "lib/addressable/uri.rb", "lib/addressable/version.rb", "spec", "spec/addressable", "spec/addressable/idna_spec.rb", "spec/addressable/net_http_compat_spec.rb", "spec/addressable/template_spec.rb", "spec/addressable/uri_spec.rb", "spec/spec_helper.rb", "st", "tasks", "tasks/clobber.rake", "tasks/gem.rake", "tasks/git.rake", "tasks/metrics.rake", "tasks/rspec.rake", "tasks/yard.rake", "website", "website/index.html"]
```
sferik: All the changes in this file were made automatically by running the `gem:gemspec` `rake` task.

.travis.yml
sporkmonger: Good point, but might be better to do a branching check to see if it's installed or not and if not, then install it, just so we don't have to go back and fix if it becomes a problem.

sferik: I’d be pretty surprised if Travis regresses in this way without warning.

sferik:
Thanks for merging!

PR #197, opened 2015-05-11
Check encoding for port

esjee:
Hey, this a patch we wrote for our local fork. We occasionally ran into this, and figured you might find it useful.


sporkmonger:
All set. Sorry about the delay.

Issue #200, opened 2015-06-01
partial expansion weirdness


pezra:
Consider the following example:

```ruby
Addressable::Template.new("http://example.com{?foo,bar}")
  .partial_expand(bar: "42").expand({}).to_s
# => "http://example.com&bar=42"
```
This results in a broken URL. It seems like the result should be `http://example.com?bar=42`. Would a PR to this effect be accepted? Would it be accepted if it resulted in a reordering of the form items when partial expansion is used? Eg,

```ruby
Addressable::Template.new("http://example.com{?foo,bar}")
  .partial_expand(bar: "42").expand({foo: 1}).to_s
# => "http://example.com?bar=42&foo=1"
```

PR #202, opened 2015-06-04
Avoid producing invalid urls for partial form-style query expansion

pezra:
When `http://example.com/{?a,b}` is partially expanded with `b: 2` the result must be `http://example.com/?b=2{&a}`. That way if the final expansion is done w/ `a` undefined the resulting URL will be valid.

If the form-style query vars are not reordered and final expansion is done w/ `a` undefined the result is
`http://example.com/&b=2` which is clearly malformed.

### DIFF
--- a/lib/addressable/template.rb
+++ b/lib/addressable/template.rb
@@ -720,19 +720,54 @@ module Addressable
     # @return [String] The expanded expression
     def transform_partial_capture(mapping, capture, processor = nil)
       _, operator, varlist = *capture.match(EXPRESSION)
-      is_first = true
-      varlist.split(',').inject('') do |acc, varspec|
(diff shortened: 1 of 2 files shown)


pezra:
This addresses issue #200.

sporkmonger:
Thanks for this PR, sorry it took me _so_ long to get to it.

faucct:
Is still an issue when expanding `b` to `nil`:

```ruby
require 'addressable'

puts Addressable::Template.new(<<~URI.chomp).partial_expand(b: nil).pattern
  http://example.com{?a,b}
URI
# => http://example.com{&a}
```

dentarg:
@faucct I think you should open a new issue, if you want the problem to be addressed

PR #203, opened 2015-06-09
Newline in hostname

EiNSTeiN-:
When %0a is present in the hostname and `.normalize` is called on the URI object, the newline is unescaped, which is not desirable.

```
irb(main):001:0> require 'addressable/uri'
 => true
irb(main):002:0> url = Addressable::URI.parse('http://abc%0adef/')
 => #<Addressable::URI:0x3ffebdcbde2c URI:http://abc%0adef/>
irb(main):003:0> url
 => #<Addressable::URI:0x3ffebdcbde2c URI:http://abc%0adef/>
irb(main):004:0> url.to_s
 => "http://abc%0adef/"
irb(main):005:0> url.normalize.to_s
 => "http://abc\ndef/"
```

ping @sporkmonger

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -1098,7 +1098,7 @@ module Addressable
 
       unreserved = CharacterClasses::UNRESERVED
       sub_delims = CharacterClasses::SUB_DELIMS
-      if @host != nil && (@host =~ /[<>{}\/\?\#\@"\n]/ ||
+      if @host != nil && (@host =~ /[<>{}\/\?\#\@"\r\n]/ ||
           (@host[/^\[(.*)\]$/, 1] != nil && @host[/^\[(.*)\]$/, 1] !~
           Regexp.new("^[#{unreserved}#{sub_delims}:]*$")))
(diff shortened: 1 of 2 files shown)


lib/addressable/uri.rb
```diff
@@ -1094,7 +1098,7 @@ def host=(new_host)
 
       unreserved = CharacterClasses::UNRESERVED
       sub_delims = CharacterClasses::SUB_DELIMS
-      if @host != nil && (@host =~ /[<>{}\/\?\#\@"]/ ||
+      if @host != nil && (@host =~ /[<>{}\/\?\#\@"\r\n]/ ||
```
ptoomey3: Maybe something like `[[:space:]]` would be more general purpose and allow us to catch more whitespace related values?

ptoomey3:
Vaguely related: https://github.com/sporkmonger/addressable/pull/205

EiNSTeiN-:
@sporkmonger, I updated this PR with your suggested changes. Sorry for the delay!

sporkmonger:
All set. Sorry about the delay.

PR #205, opened 2015-06-25
Rfc3986 scheme validation

ptoomey3:
The current version of Addressable does the correct scheme check (alpha followed by zero or more alpha,digit, or (+-.), but it [does not use any regex anchors](https://github.com/sporkmonger/addressable/blob/06896a8e3cad8719e30dfb4d0c64aee2cacd3309/lib/addressable/uri.rb#L865). So, it allows any scheme so long as some portion of it is rfc compliant. For example, the current release allows `a.scheme = "1test"` since that is rfc compliant beginning at character `1`. But, it won't allow `a.scheme = "111"` because no part of that is compliant. This PR updates the scheme validation to be more strict.

P.S. It seems like there is a fair bit of use of the `^` and `$` regex anchors. This is ok in some locations, but probably not ideal in others (I haven't audited them). I would imagine, in general, you would want to be matching on beginning/end of the string and not just on beginning/end of line.


Gemfile
```diff
@@ -9,7 +9,7 @@ end
 
 group :test, :development do
   gem 'rake', '>= 0.7.3'
-  gem 'rspec', '>= 2.9.0'
+  gem 'rspec', '~> 2.14.0'
```
ptoomey3: I changed this `rspec` version since the current Gemfile was causing it to use the latest rspec release. Unfortunately, that caused almost all of the tests to fail, as there are a number of deprecated rspec uses in the current tests. I briefly tried updating the tests, but decided this was a better short-term solution for this PR. Version 2.14 seems to be the latest in the 2.x line.

lib/addressable/uri.rb
```diff
@@ -860,7 +860,7 @@ def scheme=(new_scheme)
       elsif new_scheme
         new_scheme = new_scheme.to_str
       end
-      if new_scheme && new_scheme !~ /[a-z][a-z0-9\.\+\-]*/i
+      if new_scheme && new_scheme !~ /\A[a-z][a-z0-9\.\+\-]*\z/i
```
ptoomey3: The regex was correct; it just lacked the proper anchors.

spec/addressable/uri_spec.rb
```diff
@@ -3511,22 +3553,6 @@ def to_s
         "http://user:pass@example.com/path/to/resource?query=x#fragment")
   end
 
-  it "should have the correct scheme after assignment" do
```
ptoomey3: Doh..it looks like i dropped more of this test than I should have. I'll pull it back and just drop the specific `bogus!` test.

Gemfile
ptoomey3: Doh..I just realized I must have already had a fork of this repo that was out of date. I'm merging upstream master to catch up. Sorry about that :smile:

ptoomey3:
polite bump

sporkmonger:
All set. Sorry about the delay!

PR #212, opened 2015-10-02
Replace concatenation with extrapolation

nplusp:
it's faster 🚀


sporkmonger:
This is a micro-optimization that hurts readability and violates the 80 char max coding style this library tries to adhere to. Unless you have benchmarks demonstrating this is a major bottleneck for a realistic use-case, there's no good reason to merge this.

Issue #168, opened 2014-09-16
Require 'addressable/uri' automatically


conf:
Basically, the question is taken from this [stackoverflow question](http://stackoverflow.com/questions/12497138/addressable-gem-is-not-found), i.e. if I have `addressable` in my Gemfile
```ruby
gem 'addressable'
```
I also have to explicitly `require 'addressable/uri'` in order to use `Addressable::URI`. 
Can we get rid of this extra step? Thank you.

sporkmonger:
The main reason I did it this way was because there is also `Addressable::Template`, and most people who want `Addressable::URI` do not need `Addressable::Template`. But I suppose there's no reason you couldn't separately load them since the top-level require can be optional.

conf:
Well, maybe it will be better to move out `Addressable::Template` to its own gem and make `addressable` depend on it? This way it would be more SRP and people who want only `Addressable::Template` would use that gem only.

PR #213, opened 2015-10-08
Adding section in readme to explain how to use with Bundler

willtcarey:
I had a little bit of difficulty setting this bundle up in my Gemfile at first because I didn't realize you had split the gem into 2 separate modules that you have to explicitly include. I figured it out after a while but hopefully by putting this info in the readme it will alleviate some confusion

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,3 +103,12 @@ $ sudo apt-get install idn # Debian/Ubuntu
 $ brew install libidn # OS X
 $ gem install idn-ruby
 ```
+
+# Bundler
+
+```ruby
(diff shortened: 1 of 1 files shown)


kyanny:
@willtcarey 
It seems already resolved at master https://github.com/sporkmonger/addressable/commit/0475980bb943dfcac2141c8f40b45ea2e6e5f76f
Similar issue https://github.com/sporkmonger/addressable/issues/168

`gem "addressable", github: "sporkmonger/addressable"` would help.

willtcarey:
Ah, that will work fine. I didn't realize it was already addressed in the next release

PR #223, opened 2016-01-19
Work with ruby 2.3's --enable-frozen-string-literal

jeremyevans:
These changes are the minimal ones necessary to allow Sequel's specs
to pass. There may well be other changes that are required.

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -1151,7 +1151,7 @@ module Addressable
     # @return [String] The authority component.
     def authority
       self.host && @authority ||= begin
-        authority = ""
+        authority = String.new
         if self.userinfo != nil
           authority << "#{self.userinfo}@"
         end
@@ -1170,7 +1170,7 @@ module Addressable
     def normalized_authority
       return nil unless self.authority
       @normalized_authority ||= begin
-        authority = ""
+        authority = String.new
         if self.normalized_userinfo != nil
           authority << "#{self.normalized_userinfo}@"
         end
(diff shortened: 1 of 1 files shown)


sporkmonger:
Are there tests you can write that would fail without this change?

jeremyevans:
Have you tried running the addressable specs with `ruby --enable-frozen-string-literal` using ruby 2.3?  I'm assuming they would fail if you didn't make this change.  They may fail for other reasons in addition to this change.  To be fair, I didn't do that.  However, from testing Sequel (using data_objects, which uses addressable), this appears to be an issue.

sporkmonger:
No, I hadn't, I'll give it a go, see what happens.

sporkmonger:
Can't get it past bundler actually. How did you even run Sequel's tests? I'm getting exceptions long before it hits the test runner.

jeremyevans:
I've heard bundler's master branch (unreleased) contains fixes.  Personally, I don't use bundler for testing (or much at all), so it wasn't an issue for me.

If you just want to diagnose frozen string literal issues in your own code without worrying about other people's code, just add `# frozen-string-literal: true` to the top of all of the lib files, and don't run with `--enable-frozen-string-literal`.

sporkmonger:
Yeah, I see the test failures now.

sporkmonger:
I merged this because it seemed to have an improvement, however it definitely does not fix all failing tests.

sporkmonger:
I see 498 test failures still with frozen string literals.

PR #225, opened 2016-01-29
Test against ruby 2.3

antstorm:
Hi,

Here's a very small PR with 2 changes:

1. Add ruby 2.3.0 as a testing target for Travis
2. Lock `2.2.4` and `2.1.8` rubies, otherwise Travis is using `2.1.5` and `2.2.0` instead of the new ones (can be verified by checking out the last Travis build output)


sporkmonger:
Thanks!

PR #229, opened 2016-02-23
Remove explicit query param sorting

### LINKED ISSUE #192: Round-tripping query_values unnecessarily changes order.
~~~ ruby
puts Addressable::URI.parse('/foo?d=1&b=2&e=3&a=4&c=5').tap do |uri|
  uri.query_values = uri.query_values
end.normalize
#=> /foo?a=4&b=2&c=5&d=1&e=3
~~~

The 'problem' with this is that if I am cleaning an existing URI by removing specific query string values, the URI almost always changes, even when no actual changes occurred.

The culprit is  
https://github.com/sporkmonger/addressable/blob/master/lib/addressable/uri.rb#L1595

Which has the comment:

~~~ ruby
# Useful default for OAuth and caching.
# Only to be used for non-Array inputs. Arrays should preserve order.
new_query_values.sort!
~~~

Ruby's hashes are intentionally ordered, and with this code I cannot as-easily round-trip a URI's parameters through this and have them unchanged.

If this `sort!` is not removed wholesale, I request that an option be added to not use it, or a separate method (`uri.sorted_query_values=` ?) that does not sort.

yez:
@sporkmonger 

This fixed #192 and removes the sorting for query values.

Since this feature was used to support the now unsupported
  1.8.x, it can be removed.

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -1638,9 +1638,6 @@ module Addressable
           key = key.to_s if key.kind_of?(Symbol)
           [key, value]
         end
-        # Useful default for OAuth and caching.
-        # Only to be used for non-Array inputs. Arrays should preserve order.
-        new_query_values.sort!
       end
 
       # new_query_values have form [['key1', 'value1'], ['key2', 'value2']]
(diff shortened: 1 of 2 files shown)


sporkmonger:
Please add a semver version bump and update CHANGELOG since this is a breaking change.

yez:
Sounds good. Since this is a change in the contract change for the `query_values=` method, is the correct sem version update to 3.0?

sporkmonger:
It would be yes, so actually this should get merged to the three branch, in which case only the changelog update is necessary, now that I think about it.

yez:
Ok I will close this PR and open one against the three branch

PR #232, opened 2016-03-17
Remove duplicate code block

ieatfood:


sporkmonger:
Code coverage went down, which it shouldn't have if this was really a duplicate block. You might have mistaken it for a dupe w/ the tests immediately above or below it, but it is *not* a dupe.

walro:
Hi @sporkmonger,

I was curios on this. To me it looks like these lines are indeed duplicates: https://github.com/sporkmonger/addressable/blob/295d7baec1281976434e9061b8065979b3f38c83/spec/addressable/uri_spec.rb#L120-L134

I get your coverage point though (very reasonable) so I took a look at https://coveralls.io/github/sporkmonger/addressable and at the time of writing this it seems like the four latests build all show the same decrease in coverage (-0.4%). Unless I oversee something I must conclude that this PR was valid and that coverall has changed something on their side.

dentarg:
> Code coverage went down

Must have been a bug that code coverage went down because https://github.com/sporkmonger/addressable/pull/249 (the exact same diff) is green 😄

PR #235, opened 2016-04-18
Remove explicit query param sorting

yez:
@sporkmonger 

Re-opening https://github.com/sporkmonger/addressable/pull/229 since this is a 3.0 feature.

Running the build seems to have an issue with the `its` method but none of my code changes that, so not sure what the problem is there.

Added the change the the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Addressable 3.0
+- removed default query param sorting
+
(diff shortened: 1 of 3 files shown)


sporkmonger:
I agree w/ the logic, the code looks totally right to me, and since it's a merge to the `three` branch instead of `master` I'm OK w/ just merging without tests passing.

PR #243, opened 2016-09-23
Fix query values swallowing unescaped html

### LINKED ISSUE #242: unescaped HTML in query values
Had an issue where a url with unescaped HTML was being accepted into the query value. 
This caused an issue when the url was then rebuilt with extra queries merged as the order was sorted resulting in the queries being bad:
`?&amp;a=...`

E.g.
Input URL: `example.com?string=foo&amp;another=bar`
resulting in `uri.query_values => {"string"=>"foo", "amp;another"=>"bar"}`
when new params added the returned url looks like: 
`example.com?amp%3Banother=bar&string=foo&...`

Pull Request to follow shortly to fix this..
`uri.query_values => {"string"=>"foo", "another"=>"bar"}`
`example.com?another=bar&string=foo&...`

willhall88:
resolves and closes #242

Input URL: `example.com?string=foo&amp;another=bar`

BEFORE:
`uri.query_values => {"string"=>"foo", "amp;another"=>"bar"}`
...
`example.com?amp%3Banother=bar&string=foo&...`

AFTER:
`uri.query_values => {"string"=>"foo", "another"=>"bar"}`
...
`example.com?another=bar&string=foo&...`

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -1546,7 +1546,7 @@ module Addressable
       if new_query && !new_query.respond_to?(:to_str)
         raise TypeError, "Can't convert #{new_query.class} into String."
       end
-      @query = new_query ? new_query.to_str : nil
+      @query = new_query ? CGI.unescapeHTML(new_query.to_str) : nil
 
       # Reset dependent values
       remove_instance_variable(:@normalized_query) if defined?(@normalized_query)
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -18,6 +18,7 @@
 
 require "addressable/version"
 require "addressable/idna"
+require "cgi"
 
 ##
 # Addressable is a library for processing links and URIs.
--- a/spec/addressable/uri_spec.rb
+++ b/spec/addressable/uri_spec.rb
@@ -4132,6 +4132,25 @@ describe Addressable::URI, "when parsed from " +
   end
 end
 
+describe Addressable::URI, "when parsed from " \
+                           "'http://example.com/?q=a&amp;q2=b'" do
+  before do
+    @uri = Addressable::URI.parse("http://example.com/?q=a&amp;q2=b")
+  end
+
+  it "should have a query of 'q=a&q2=b'" do
+    expect(@uri.query).to eq("q=a&q2=b")
+  end
+
+  it "should have query_values of {'q' => 'a', 'q2' => 'b'}" do
+    expect(@uri.query_values).to eq("q" => "a", "q2" => "b")
+  end
+
+  it "should have a normalized query of 'q=a&q2=b'" do
+    expect(@uri.normalized_query).to eq("q=a&q2=b")
+  end
+end
+
 describe Addressable::URI, "when parsed from " +
     "'http://example.com/?v=%7E&w=%&x=%25&y=%2B&z=C%CC%A7'" do
   before do


dentarg:
Not sure why CI (2.2.4 and 2.3.0) pass, but I think you are missing `require "cgi"`:

```ruby
$ git co -t willhall88/fix-unescaped-queries
Branch fix-unescaped-queries set up to track remote branch fix-unescaped-queries from willhall88 by rebasing.
Switched to a new branch 'fix-unescaped-queries'

~/src/addressable fix-unescaped-queries
$ pry
[1] pry(main)> $: << '../lib' << 'lib'
=> ["/Users/dentarg/.gem/ruby/2.2.5/gems/slop-3.6.0/lib",
 "/Users/dentarg/.gem/ruby/2.2.5/gems/method_source-0.8.2/lib",
 "/Users/dentarg/.gem/ruby/2.2.5/gems/pry-0.10.4/lib",
 "/Users/dentarg/.gem/ruby/2.2.5/gems/coderay-1.1.1/lib",
 "/Users/dentarg/.gem/ruby/2.2.5/gems/byebug-9.0.5/lib",
 "/Users/dentarg/.gem/ruby/2.2.5/extensions/x86_64-darwin-15/2.2.0-static/byebug-9.0.5",
 "/Users/dentarg/.gem/ruby/2.2.5/gems/pry-byebug-3.4.0/lib",
 "/Users/dentarg/.gem/ruby/2.2.5/gems/psych-2.1.0/lib",
 "/Users/dentarg/.gem/ruby/2.2.5/extensions/x86_64-darwin-15/2.2.0-static/psych-2.1.0",
 "/Users/dentarg/.gem/ruby/2.2.5/gems/pry-doc-0.9.0/lib",
 "/Users/dentarg/.gem/ruby/2.2.5/gems/yard-0.9.2/lib",
 "/Users/dentarg/.rubies/ruby-2.2.5/lib/ruby/site_ruby/2.2.0",
 "/Users/dentarg/.rubies/ruby-2.2.5/lib/ruby/site_ruby/2.2.0/x86_64-darwin15",
 "/Users/dentarg/.rubies/ruby-2.2.5/lib/ruby/site_ruby",
 "/Users/dentarg/.rubies/ruby-2.2.5/lib/ruby/vendor_ruby/2.2.0",
 "/Users/dentarg/.rubies/ruby-2.2.5/lib/ruby/vendor_ruby/2.2.0/x86_64-darwin15",
 "/Users/dentarg/.rubies/ruby-2.2.5/lib/ruby/vendor_ruby",
 "/Users/dentarg/.rubies/ruby-2.2.5/lib/ruby/2.2.0",
 "/Users/dentarg/.rubies/ruby-2.2.5/lib/ruby/2.2.0/x86_64-darwin15",
 "../lib",
 "lib"]
[2] pry(main)> require 'addressable/uri'
=> true
[3] pry(main)> uri = Addressable::URI.parse("http://example.com/?q=a&amp;q2=b")
NameError: uninitialized constant Addressable::URI::CGI
from /Users/dentarg/src/addressable/lib/addressable/uri.rb:1549:in `query='
```

dentarg:
Also, I think your approach doesn't work.


```ruby
[7] pry(main)> uri1 = Addressable::URI.parse("http://example.com/?q=a&amp;q2=b")
=> #<Addressable::URI:0x3fcd824a6430 URI:http://example.com/?q=a&q2=b>
[8] pry(main)> uri2 = Addressable::URI.parse("http://example.com/?q=a&#111;q2=b")
=> #<Addressable::URI:0x3fcd83ed1918 URI:http://example.com/?q=a&#111;q2=b>
[9] pry(main)> uri1.query_values
=> {"q"=>"a", "q2"=>"b"}
[10] pry(main)> uri2.query_values
=> {"q"=>"a"}
```

Is this expected?

dentarg:
Oops, sorry... :) master gives the same thing:

```
[3] pry(main)> uri2 = Addressable::URI.parse("http://example.com/?q=a&#111;q2=b")
=> #<Addressable::URI:0x3fc2ffd018d8 URI:http://example.com/?q=a&#111;q2=b>
[4] pry(main)> uri2.query_values
=> {"q"=>"a"}
```

dentarg:
Okay, this is a better example (my bad using something including the anchor (`#`))

Your branch:

```ruby
[12] pry(main)> uri2 = Addressable::URI.parse("http://example.com/?q=a&gt;q2=b")
=> #<Addressable::URI:0x3fcd82d35434 URI:http://example.com/?q=a>q2=b>
[13] pry(main)> uri2.query_values
=> {"q"=>"a>q2=b"}
```

current master (a15b704):

```ruby
[7] pry(main)> uri2 = Addressable::URI.parse("http://example.com/?q=a&gt;q2=b")
=> #<Addressable::URI:0x3fc2fe0b8fc8 URI:http://example.com/?q=a&gt;q2=b>
[8] pry(main)> uri2.query_values
=> {"q"=>"a", "gt;q2"=>"b"}
```

My point is... is [the spec](https://tools.ietf.org/html/rfc3986) telling us to do this unescaping?

willhall88:
> Not sure why CI (2.2.4 and 2.3.0) pass, but I think you are missing require "cgi"

Yeah, good shout. It works fine on my machine without needing to require it, but I can put this in.

willhall88:
I get exactly the same behaviour on both branches when I run the same inputs...

`fix-unescaped-queries` branch:
```
[1] pry(main)> require 'addressable/uri'
=> true
[2] pry(main)> uri2 = Addressable::URI.parse("http://example.com/?q=a&gt;q2=b")
=> #<Addressable::URI:0x3fdc81de4cb4 URI:http://example.com/?q=a&gt;q2=b>
[3] pry(main)> uri2.query_values
=> {"q"=>"a", "gt;q2"=>"b"}

```
current master (a15b704):

```
[1] pry(main)> require 'addressable/uri'
=> true
[2] pry(main)> uri2 = Addressable::URI.parse("http://example.com/?q=a&gt;q2=b")
=> #<Addressable::URI:0x3ffd5a23bf6c URI:http://example.com/?q=a&gt;q2=b>
[3] pry(main)> uri2.query_values
=> {"q"=>"a", "gt;q2"=>"b"}
```

dentarg:
> On 26 sep. 2016, at 11:06, Will Hall <notifications@github.com> wrote:
> 
> I get exactly the same behaviour on both branches when I run the same inputs...

Are you sure you are loading the code from your branch? Looks like your are requiring from ruby gems

dentarg:
See my first comment how I changed the load path with

    $: << '../lib' << 'lib'

willhall88:
Ah yeah my bad, sorry, I see what you're saying now. :) 

So there's at least two alternative options instead of unescaping the html..

Currently when finding the query values, the string is split on `&` on [line 1582](https://github.com/sporkmonger/addressable/blob/master/lib/addressable/uri.rb#L1582)
This should maybe only split on `&` or `&amp;`, but not on any other html entities (i.e. not `&gt;`).. I feel like this would involve some sort of regex manipulation.. perhaps not the best. 

The issue itself #242, comes about because the original queries get sorted when merging new queries. A potential alternative fix for the issue would be to stop the original queries from being sorted and keeping the original order. Again although it would fix the issue I've seen, it perhaps isn't the best solution as there would still be cases where the query keys would still be contaminated with parts of html entities i.e. `"gt;q2"=>"b"` or `"amp;q2"=>"b"`

What are your thoughts?

sporkmonger:
So before anyone gets too much further in on this, I explicitly won't be accepting this PR. HTML entity encoding/unencoding simply does not belong in a URI parsing library. You're more than welcome to run a fork but I would recommend moving code to resolve this issue to someplace that runs prior to parsing URIs.

PR #244, opened 2016-09-29
Stopped merge!, join!, omit! and normalize! from disabling deferred validation

iainbeeston:
In addressable 2.4 there is a bug, where if you defer validation, then
use merge!, join!, omit! or normalize! inside the block, deferred
validation is disabled immediately and all further modification to the
uri inside the block is validated immediately.

This is happening because those methods all call replace_self, which
removes all instance variables, including the one used to flag whether
validation is deferred or not. Once this has been removed, all calls to
validate are actioned immediately.

This bug causes a performance penalty, and also lots of ruby
warnings ("warning: instance variable @validation_deferred not
initialized").

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- stopped merge!, join!, omit! and normalize! from disabling deferred validation
+
 # Addressable 2.4.0
 - support for 1.8.x dropped
 - double quotes in a host now raises an error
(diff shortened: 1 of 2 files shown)


iainbeeston:
@sporkmonger Is there any chance we can get this merged?

sporkmonger:
Definitely.

PR #245, opened 2016-09-29
Fix travis build

iainbeeston:
The travis build is failing on many platforms at the moment. Mostly this is because some dependent gems no longer support older versions of ruby. This is my attempt to fix the build so other pull requests can be merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ matrix:
   allow_failures:
     - rvm: ruby-head
     - rvm: jruby-head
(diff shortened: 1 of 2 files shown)


iainbeeston:
FYI I've put in code to check `RUBY_VERSION` as the `platforms` directive in bundler doesn't really do anything in bundler < 1.14

sporkmonger:
Thanks. Change seems totally reasonable to me.

PR #256, opened 2017-02-03
Add information for some usefull functions

rst0git:


sporkmonger:
The README doesn't need to be complete documentation for the library. I'd be happy to see this go into the Wiki though.

Issue #258, opened 2017-03-24
Addressable::Template#expand transforms some multi-byte string to another unintended string.


civitaspo:
Hi maintainers,

I use `addressable` in google-api-client-ruby. I found the below bug, so could you fix it?

```
irb(main):001:0> require 'addressable'
=> true
irb(main):002:0> url = Addressable::Template.new('https://sheets.googleapis.com/v4/spreadsheets/{spreadsheetId}/values/{range}')
=> #<Addressable::Template:0x7d0 PATTERN:https://sheets.googleapis.com/v4/spreadsheets/{spreadsheetId}/values/{range}>
irb(main):003:0> url.expand(spreadsheetId: 'test', range: "日本語（）")
=> #<Addressable::URI:0x7d2 URI:https://sheets.googleapis.com/v4/spreadsheets/test/values/%E6%97%A5%E6%9C%AC%E8%AA%9E%28%29>
irb(main):004:0> URI.unescape('https://sheets.googleapis.com/v4/spreadsheets/test/values/%E6%97%A5%E6%9C%AC%E8%AA%9E%28%29')
=> "https://sheets.googleapis.com/v4/spreadsheets/test/values/日本語()"
## why is this not 'https://sheets.googleapis.com/v4/spreadsheets/test/values/%e6%97%a5%e6%9c%ac%e8%aa%9e%ef%bc%88%ef%bc%89' ?
```

`Addressable::Template#expand` transforms multi-byte brackets `（）` to single-byte brackets `()`.

civitaspo:
Oh, I forgot writing my environment.

addressable version: 2.5.0
ruby version: jruby 9.1.8.0 (2.3.1) 2017-03-06 90fc7ab Java HotSpot(TM) 64-Bit Server VM 25.121-b13 on 1.8.0_121-b13 +jit [darwin-x86_64]

civitaspo:
This code https://github.com/sporkmonger/addressable/blob/4de11a4b95580607f66e7e1829e5aaeacb07a086/lib/addressable/idna/pure.rb#L114-L121 intentionally transforms multi-byte string to single-byte string if possible.

sporkmonger:
As mentioned in #259, it's actually required by the URI Template spec. We're going to make it optional though.

sporkmonger:
2.5.1 released w/ optional normalization.

civitaspo:
thx so much!

PR #259, opened 2017-03-29
Support byte semantics for templates.

sqrrrl:
Addressable is overzealous in interpreting the RFC. Normalization of unicode in template expansion is recommended for user-supplied values only. Addressable shouldn't make any assumptions about whether the value needs normalization -- it's better left to the calling application to decide.

This came up wrt GoogleCloudPlatform/google-cloud-ruby#1364 where Cloud Storage uses byte semantics for resource names. The forced normalization in this library results in incorrect URIs for resources that were created with a different encoding.

This PR keeps the default behavior, but adds an option to skip normalization.

### DIFF
--- a/lib/addressable/template.rb
+++ b/lib/addressable/template.rb
@@ -488,6 +488,8 @@ module Addressable
     # @param [Hash] mapping The mapping that corresponds to the pattern.
     # @param [#validate, #transform] processor
     #   An optional processor object may be supplied.
+    # @param [Boolean] normalize_values
+    #   Optional flag to enable/disable unicode normalization. Default: true
     #
     # The object should respond to either the <tt>validate</tt> or
     # <tt>transform</tt> messages or both. Both the <tt>validate</tt> and
@@ -518,7 +520,7 @@ module Addressable
     #     "http://example.com/{?one,two,three}/"
     #   ).partial_expand({"one" => "1", "three" => 3}).pattern
     #   #=> "http://example.com/?one=1{&two}&three=3"
-    def partial_expand(mapping, processor=nil)
+    def partial_expand(mapping, processor=nil, normalize_values=true)
       result = self.pattern.dup
       mapping = normalize_keys(mapping)
       result.gsub!( EXPRESSION ) do |capture|
@@ -533,6 +535,8 @@ module Addressable
     # @param [Hash] mapping The mapping that corresponds to the pattern.
     # @param [#validate, #transform] processor
     #   An optional processor object may be supplied.
+    # @param [Boolean] normalize_values
+    #   Optional flag to enable/disable unicode normalization. Default: true
     #
     # The object should respond to either the <tt>validate</tt> or
     # <tt>transform</tt> messages or both. Both the <tt>validate</tt> and
@@ -583,11 +587,11 @@ module Addressable
     #     ExampleProcessor
     #   ).to_str
     #   #=> Addressable::Template::InvalidTemplateValueError
-    def expand(mapping, processor=nil)
+    def expand(mapping, processor=nil, normalize_values=true)
       result = self.pattern.dup
       mapping = normalize_keys(mapping)
       result.gsub!( EXPRESSION ) do |capture|
-        transform_capture(mapping, capture, processor)
+        transform_capture(mapping, capture, processor, normalize_values)
       end
       return Addressable::URI.parse(result)
     end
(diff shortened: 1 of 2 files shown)


sqrrrl:
Also related to #258

spec/addressable/template_spec.rb
```diff
@@ -1013,6 +1013,22 @@ def self.match(name)
     end
   end
   describe "Expand" do
+    context "expand with unicode values" do
+      subject {
```
sporkmonger: Agree w/ Hound.

spec/addressable/template_spec.rb
```diff
@@ -1013,6 +1013,22 @@ def self.match(name)
...
+        Addressable::Template.new("http://example.com/search/{query}/")
+      }
+      it "normalizes unicode by default" do
+        expect(subject.expand({"query" => "Cafe\u0301"}).to_str).to eq(
```
sporkmonger: Fine to ignore.

spec/addressable/template_spec.rb
```diff
@@ -1013,6 +1013,22 @@ def self.match(name)
...
+          "http://example.com/search/Caf%C3%A9/"
+        )
+      end
+
+      it "does not normalize unicode when byte semantics requested" do
+        expect(subject.expand({"query" => "Cafe\u0301"}, nil, false).to_str).to eq(
```
sporkmonger: Please line-wrap, but ignore spacing comment.

lib/addressable/template.rb
```diff
@@ -518,7 +520,7 @@ def match(uri, processor=nil)
     #     "http://example.com/{?one,two,three}/"
     #   ).partial_expand({"one" => "1", "three" => 3}).pattern
     #   #=> "http://example.com/?one=1{&two}&three=3"
-    def partial_expand(mapping, processor=nil)
+    def partial_expand(mapping, processor=nil, normalize_values=true)
```
sporkmonger: I think hound's detection of unused method argument is incorrect? Ignore this.

lib/addressable/template.rb
```diff
@@ -877,7 +884,7 @@ def transform_capture(mapping, capture, processor=nil)
             end
             if processor.respond_to?(:transform)
               transformed_value = processor.transform(name, value)
-              transformed_value = normalize_value(transformed_value)
+              transformed_value = normalize_value(transformed_value) if normalize_values
```
sporkmonger: Given the line length, I'd put it in a block-if.

lib/addressable/template.rb
```diff
@@ -583,11 +587,11 @@ def partial_expand(mapping, processor=nil)
     #     ExampleProcessor
     #   ).to_str
     #   #=> Addressable::Template::InvalidTemplateValueError
-    def expand(mapping, processor=nil)
+    def expand(mapping, processor=nil, normalize_values=true)
```
sporkmonger: Ignore.

lib/addressable/template.rb
sporkmonger: It does seem like the parameter isn't getting used. Possible code omission?

lib/addressable/template.rb
```diff
@@ -794,7 +801,7 @@ def operator_sequence(operator)
     # after sending the value to the transform method.
     #
     # @return [String] The expanded expression
-    def transform_capture(mapping, capture, processor=nil)
+    def transform_capture(mapping, capture, processor=nil, normalize_values=true)
```
sporkmonger: I know, 1 over, but please line wrap.

adstage-david:
While I'm not sure the template expansion is being overzealous in this scenario, I'm definitely fine with making it optional to disable to allow more use cases. Probably do want to handle the partial expand case too (which falls through to the transform_capture method already fixed)?

sqrrrl:
Fixed the partial_expand casae to match.

sqrrrl:
One more change coming to fix style issues

blowmage:
@sporkmonger Assuming the the code formatting issues are addressed, is this a change that you would consider accepting?

sporkmonger:
@blowmage Yes.

spec/addressable/template_spec.rb
```diff
@@ -969,6 +969,22 @@ def self.match(name)
         )
       end
     end
+    context "partial expand with unicode values" do
+      subject {
```
sporkmonger: `do .. end`

spec/addressable/template_spec.rb
sporkmonger: `do .. end`

sporkmonger:
I think the build may have hung. Going to merge, tests already passed in all required environments, and it'll rerun once it hits `master` anyway.

sporkmonger:
Released.

PR #260, opened 2017-07-02
Get tests passing with frozen string literals.

pat:
These changes allow Addressable to be used when Ruby's newish frozen string literal feature is enabled (via `RUBYOPT="--enable-frozen-string-literal" when running MRI 2.4.x).

The test suite passes when I have this flag set _and_ I'm using the latest in the git repos for:

* rspec
* rspec-core
* rspec-support
* rspec-expectations
* rspec-mocks
* simplecov
* simplecov-html

I've not updated the Gemfile to include these repos, nor the Travis CI configuration to use the RUBYOPT flag for MRI 2.4+, but I can if you'd like? :)

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -436,7 +436,7 @@ module Addressable
       uri = uri.dup
       # Seriously, only use UTF-8. I'm really not kidding!
       uri.force_encoding("utf-8")
-      leave_encoded.force_encoding("utf-8")
+      leave_encoded = leave_encoded.dup.force_encoding("utf-8")
       result = uri.gsub(/%[0-9a-f]{2}/iu) do |sequence|
         c = sequence[1..3].to_i(16).chr
         c.force_encoding("utf-8")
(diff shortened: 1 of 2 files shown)


lib/addressable/uri.rb
```diff
@@ -1032,9 +1032,9 @@ def normalized_userinfo
         if !current_user && !current_password
           nil
         elsif current_user && current_password
-          "#{current_user}:#{current_password}"
+          "#{current_user}:#{current_password}".dup
         elsif current_user && !current_password
-          "#{current_user}"
+          "#{current_user}".dup
```
sporkmonger: Yeah, this line looks odd to me. I probably wrote it that way absent-mindedly, but this code just seems _bad_ to me looking back on it.

sporkmonger:
Calling `dup` on string literals strikes me as a low-readability solution to this problem. I'm really not super fond of it. If it's the only solution, OK, but would prefer an alternative if one exists. I haven't really spent on time on this problem though, so I really don't know.

pat:
Yeah, the `dup` syntax is really not ideal. I've opted for that in most cases because it's what the Rails core team prefer, but another option is to use `String.new("")`. I've not come across anything more elegant thus far, I'm afraid.

If encoding matters (and it certainly does in most cases), then `String.new()` should not be used, as if there's no argument the encoding is ASCII. Whereas when there is an argument, its encoding is passed through to the outer object.

Beyond that, an alternative is to use `+=` instead of `<<`, but that means many strings are allocated to memory, which is not ideal - and doesn't cover off other situations anyway.

sporkmonger:
Alright, that's a convincing argument.

PR #262, opened 2017-07-04
Request URI methods: handle URI with uppercase letters in scheme

### LINKED ISSUE #261: wrong request_uri for capitalized http urls
example:
```
uri = Addressable::URI.parse( 'http://somehost.com/somepath')
raise if uri.request_uri.nil?
uri = Addressable::URI.parse( 'HTTP://SOMEHOST.COM/SOMEPATH')
raise if uri.request_uri.nil?
```
will cause unhandled exception on last string

dentarg:
I think this is in line with https://tools.ietf.org/html/rfc3986#section-3.1

Close #261

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -1724,7 +1724,7 @@ module Addressable
     #
     # @return [String] The request URI required for an HTTP request.
     def request_uri
-      return nil if self.absolute? && self.scheme !~ /^https?$/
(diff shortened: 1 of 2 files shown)


dentarg:
Not sure I should address the comments from @houndci-bot

sporkmonger:
Yeah, I don't care about the specific nitpicks that Hound's dropping here.

sporkmonger:
Thanks for the PR!

PR #267, opened 2017-08-04
Allow public_suffix 3

twalpole:
Allow the use of public_suffix 3.x with addressable

### DIFF
--- a/gemfiles/Gemfile.public_suffix_2
+++ b/gemfiles/Gemfile.public_suffix_2
@@ -1,0 +1,34 @@
+source 'https://rubygems.org'
+
+gemspec path: ".."
+
+gem 'public_suffix', '>= 2.0.2', '~> 2.0'
+
+group :test do
+  gem 'rspec', '~> 3.0'
+  gem 'rspec-its', '~> 1.1'
+end
+
+group :development do
+  gem 'launchy', '~> 2.4', '>= 2.4.3'
+  gem 'redcarpet', :platform => :mri_19
+  gem 'yard'
+end
+
+group :test, :development do
+  gem 'rake', '> 10.0', '< 12'
+  gem 'simplecov', :require => false
+  gem 'coveralls', :require => false, :platforms => [
+    :ruby_20, :ruby_21, :ruby_22, :ruby_23
+  ]
+  # Used to test compatibility.
+  gem 'rack-mount', git: 'https://github.com/sporkmonger/rack-mount.git', require: 'rack/mount'
+
+  if RUBY_VERSION.start_with?('2.0', '2.1')
+    gem 'rack', '< 2', :require => false
+  else
+    gem 'rack', :require => false
+  end
+end
+
+gem 'idn-ruby', :platform => [:mri_20, :mri_21, :mri_22]
--- a/tasks/gem.rake
+++ b/tasks/gem.rake
@@ -20,7 +20,7 @@ namespace :gem do
 
     s.required_ruby_version = '>= 2.0'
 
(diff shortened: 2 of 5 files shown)


tasks/gem.rake
```diff
@@ -20,7 +20,7 @@ namespace :gem do
 
     s.required_ruby_version = '>= 2.0'
 
-    s.add_runtime_dependency 'public_suffix', '~> 2.0', '>= 2.0.2'
+    s.add_runtime_dependency 'public_suffix', '>= 2.0.2', '< 4.0'
```
jesperronn: @houndci-bot should be consistent to all lines (and not only the affected ones), and is irrelevant for the change here.

(Yes I know you're a bot)

jesperronn:
background info found in https://github.com/weppos/publicsuffix-ruby/blob/master/CHANGELOG.md#release-300 and https://github.com/weppos/publicsuffix-ruby/pull/133 

This was a major release in `public_suffix`, but if its not affecting any usages in this library, then it's probably only a minor release of addressable? @sporkmonger what do you think?

dentarg:
@jesperronn  3.0.0 of `public_suffix` dropped support for Ruby 2.0, but Addressable 
 still support it: https://github.com/sporkmonger/addressable/blob/addressable-2.5.1/.travis.yml#L9.

[Not sure](https://github.com/weppos/publicsuffix-ruby/issues/127) if that should warrant a bigger bump.

dentarg:
Ah, it doesn't matter for this change. Ruby 2.0 will use Public Suffix 2.0.5 (https://travis-ci.org/sporkmonger/addressable/jobs/261162023).

sporkmonger:
If `public_suffix` is following semver and the only breaking change is that they stopped supporting 2.0 then I'm fine w/ taking the approach of allowing either version under the assumption that any new breaking change will result in a 4.0 version.

tasks/gem.rake
sporkmonger: Yeah, at some point I'll probably do a big code style audit. No worries for this PR.

twalpole:
@sporkmonger There were other breaking changes around PublicSuffix::List, but not in any methods that addressable uses.

gemfiles/Gemfile.public_suffix_2
```diff
@@ -0,0 +1,34 @@
+source 'https://rubygems.org'
```
sporkmonger: Just to clarify, this `Gemfile` is only for testing purposes? If so, would it make more sense to name it in such a way that that's obviously, or perhaps include a comment to that effect?

sporkmonger:
OK, that should be OK. We have decent test coverage, so it ought to be alright.

gemfiles/Gemfile.public_suffix_2
twalpole: @sporkmonger It's for dev or test, just like the main /Gemfile (which isn't used when building the gem, the gemspec is) which doesn't have comments about that.

sporkmonger:
Alright cool. Might rearrange things a bit later when I have time, but seems fine for now.

farwayer:
Any chance new version will be released soon? ruby-addressable is broken in [archlinux](https://www.archlinux.org/packages/community/any/ruby-addressable/) now because public_suffix [was updated to 3.0](https://www.archlinux.org/packages/community/any/ruby-public_suffix/) in repos 😞

sporkmonger:
Done.

farwayer:
Thank you!

ybart:
Nice! Thanks 😀

PR #275, opened 2017-11-10
Addressable::URI.heuristic_parse throws exception when uri starts with a digit and has specified port

lagunoff:
Reproducing the bug:

```
2.3.1 :001 > require 'addressable'
2.3.1 :002 > Addressable::URI.heuristic_parse('123.example.com:8080')
Addressable::URI::InvalidURIError: Invalid scheme format: 123.example.com
        from /home/vlad/.rvm/gems/ruby-2.3.1/gems/addressable-2.5.2/lib/addressable/uri.rb:874:in `scheme='
        from /home/vlad/.rvm/gems/ruby-2.3.1/gems/addressable-2.5.2/lib/addressable/uri.rb:799:in `block in initialize'
        from /home/vlad/.rvm/gems/ruby-2.3.1/gems/addressable-2.5.2/lib/addressable/uri.rb:2355:in `defer_validation'
        from /home/vlad/.rvm/gems/ruby-2.3.1/gems/addressable-2.5.2/lib/addressable/uri.rb:796:in `initialize'
        from /home/vlad/.rvm/gems/ruby-2.3.1/gems/addressable-2.5.2/lib/addressable/uri.rb:136:in `new'
        from /home/vlad/.rvm/gems/ruby-2.3.1/gems/addressable-2.5.2/lib/addressable/uri.rb:136:in `parse'
        from /home/vlad/.rvm/gems/ruby-2.3.1/gems/addressable-2.5.2/lib/addressable/uri.rb:207:in `heuristic_parse'
        from (irb):4
        from /home/vlad/.rvm/rubies/ruby-2.3.1/bin/irb:11:in `<main>'
```
That happens because [heuristic_parse](https://github.com/sporkmonger/addressable/blob/master/lib/addressable/uri.rb#L207-L210)  first tries to parse `uri` and then it looks if `parser.scheme` has invalid characters, if so it tries to parse once again the same `uri` with prefixed sheme. The fix is essentially to add sheme prefix before first parse if `uri` doesn't have `://` as substring

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -194,7 +194,7 @@ module Addressable
         uri.gsub!(/^file:\/+/, "file:///")
       when uri =~ /^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/
         uri.gsub!(/^/, hints[:scheme] + "://")
-      when uri !~ /:\/+/ && hints.key?(:scheme)
+      when uri !~ /:\/+/
         uri.gsub!(/^/, hints[:scheme] + "://")
       end
       match = uri.match(URIREGEX)
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -181,18 +181,20 @@ module Addressable
       hints = {
         :scheme => "http"
       }.merge(hints)
-      case uri
-      when /^http:\/+/
+      case 
+      when uri =~ /^http:\/+/
         uri.gsub!(/^http:\/+/, "http://")
-      when /^https:\/+/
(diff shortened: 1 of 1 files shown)


dentarg:
@lagunoff Any test case we can add for this?

lagunoff:
@dentarg tests should check that Addressable::URI.heuristic_parse doesn't throw when called with uri like this
```
Addressable::URI.heuristic_parse('123.example.com:8080')
```

Mifrill:
resolved #296

lagunoff:
closing this

PR #280, opened 2017-11-22
accept more types of file: URI

### LINKED ISSUE #279: heuristic_parse breaks 'file:' URIs with UNC paths
File URIs can have 0, 1, 2, 3, 4, or sometimes 5 slashes after the `"file:"` part. `heuristic_parse` assumes always 0 or 3.

phluid61:
* "file:////share/path/file" => UNC/SMB share
* "file://localhost/path"    => "file:///path"
* "file://foo/bar"           => typo, becomes "file:///foo/bar"

Includes additional rspec tests.

Fixes #279

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -190,8 +190,12 @@ module Addressable
         uri.gsub!(/^feed:\/+http:\/+/, "feed:http://")
       when /^feed:\/+/
         uri.gsub!(/^feed:\/+/, "feed://")
-      when /^file:\/+/
-        uri.gsub!(/^file:\/+/, "file:///")
+      when %r[^file:/{4}]i
+        uri.sub!(%r[^file:/+]i, "file:////")
+      when %r[^file://localhost/]i
+        uri.sub!(%r[^file://localhost/+]i, "file:///")
+      when %r[^file:/+]i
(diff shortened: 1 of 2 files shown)


phluid61:
I can easily change it to use `//` literals and/or remove the `i` modifiers and/or use `gsub!` if that makes it more acceptable.

sporkmonger:
If anything, I'd rather convert everything to match `%r` style. It's the right approach here.

olleolleolle:
@sporkmonger Oh, if you didn’t see that - this PR now uses your recommendation of delimiters for regex.

sporkmonger:
So we recently actually got a real spec for `file:` URIs instead of a de facto standard established by common practice. I haven't had a chance to dive into it yet, but I'm inclined to wait on this PR until myself or another maintainer has had a chance to read it through.

https://tools.ietf.org/html/rfc8089

phluid61:
(Guess who authored RFC 8089...)

sporkmonger:
Bahahaha, nice. Totally missed that, LOL.

sporkmonger:
Gave it a second quick review, and I'm happy w/ where this is at, and I'll trust that you did the right things where the spec is concerned. 😁

PR #283, opened 2017-12-26
remove deprecated warning

takkanm:
BigDecimal.new is deprecated in Ruby2.5

```
spec/addressable/template_spec.rb:85: warning: BigDecimal.new is deprecated; use Kernel.BigDecimal method instead.
```

### DIFF
--- a/spec/addressable/template_spec.rb
+++ b/spec/addressable/template_spec.rb
@@ -82,7 +82,7 @@ describe "Type conversion" do
       :hello => 1234,
       :nothing => nil,
       :sym => :symbolic,
-      :decimal => BigDecimal.new('1')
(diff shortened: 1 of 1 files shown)


sporkmonger:
This needs to be backwards compatible. Do you know how far back `Kernel.BigDecimal` is available?

olleolleolle:
@sporkmonger Seems like it has been in since 1.9.3 at least: the docs bear that out, at least. http://ruby-doc.org/stdlib-1.9.3/libdoc/bigdecimal/rdoc/BigDecimal.html

sporkmonger:
OK, cool. We only support 2.x, so should be fine to just merge as-is.

Issue #282, opened 2017-11-22
Addressable wrong query_values when parsing query with array params


PetrKaleta:
Please have a look on the following example

```ruby
uri = Addressable::URI.parse 'http://foo.com/bar?par[]=1&par[]=2&par[]=3'
uri.query_values # => {"par[]"=>"3"}
```

expected query value is `{"par": ["1", "2", "3"]}`

Also this is very strange:

```ruby
uri = Addressable::URI.parse 'http://foo.com/bar'
uri.query_values = {par: [1, 2, 3]}
uri.to_s # => "http://foo.com/bar?par=1&par=2&par=3"

uri = Addressable::URI.parse uri.to_s
uri.query_values # => {"par"=>"3"}
```

Which is really inconsistent...

dentarg:
I think the Stack Overflow thread [Is array syntax using square brackets in URL query strings valid?](https://stackoverflow.com/questions/11490326/is-array-syntax-using-square-brackets-in-url-query-strings-valid) might help explain the lack of support in Addressable. The accepted answer starts out with 

> The answer is not simple.


Had to check myself, and percent encoding doesn't work either:

```ruby
$ irb -raddressable/uri
irb(main):001:0> Addressable::VERSION::STRING
=> "2.5.2"
irb(main):002:0> uri = Addressable::URI.parse("http://foo.com/?abc%5B%5D=123&abc%5B%5D=456")
=> #<Addressable::URI:0x3fe170044e30 URI:http://foo.com/?abc%5B%5D=123&abc%5B%5D=456>
irb(main):003:0> uri.query_values
=> {"abc[]"=>"456"}
irb(main):004:0> uri.to_s
=> "http://foo.com/?abc%5B%5D=123&abc%5B%5D=456"
irb(main):005:0> Addressable::URI.parse(uri.to_s)
=> #<Addressable::URI:0x3fe170031790 URI:http://foo.com/?abc%5B%5D=123&abc%5B%5D=456>
irb(main):006:0> Addressable::URI.parse(uri.to_s).query_values
=> {"abc[]"=>"456"}
```

PetrKaleta:
Yes, percent doesn't work either and btw it will be enough for me, to make it work...

PetrKaleta:
I think that also when setting a given query param to array value, it should turn to `%5B%5D` encoded form. So it **should** work like this:

```ruby
uri = Addressable::URI.parse 'http://foo.com/'
uri.query_values abc: [123, 456]
[...]

PR #286, opened 2018-01-21
Handle arrays in query params

philihp:
As noted in #282 by @PetrKaleta there's some weird behavior when passing in array params.

```ruby
uri = Addressable::URI.parse 'http://foo.com/'
uri.query_values = { 'abc' => ['123', '456'] }
uri.to_s
# => http://foo.com/?abc=123&abc=456
```

And parsing that back returns

```ruby
Addressable::URI.parse(uri.to_s).query_values
# => {"abc"=>"456"}
```

Perhaps instead it should give the URL

```ruby
# => http://foo.com/?abc%5B%5D=123&abc%5B%5D=456
```

Because parsing that out returns what was sent in

```ruby
Addressable::URI.parse(uri.to_s).query_values
# => {"abc"=>["123", "456"]}
```

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -1645,7 +1645,13 @@ module Addressable
           pair[1] = URI.unencode_component(pair[1].to_str.gsub(/\+/, " "))
         end
         if return_type == Hash
-          accu[pair[0]] = pair[1]
+          if pair[0].end_with? '[]'
+            pair[0].chomp! '[]'
+            accu[pair[0]] = [] unless accu.key?(pair[0])
+            accu[pair[0]] << pair[1]
+          else
+            accu[pair[0]] = pair[1]
+          end
         else
           accu << pair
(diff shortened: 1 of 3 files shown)


lib/addressable/uri.rb
```diff
@@ -1700,19 +1709,19 @@ def query_values=(new_query_values)
...
         if value == nil
-          buffer << "#{encoded_key}&"
+          buffer << encoded_key << "&"
         elsif value.kind_of?(Array)
           value.each do |sub_value|
             encoded_value = URI.encode_component(
               sub_value, CharacterClasses::UNRESERVED
             )
-            buffer << "#{encoded_key}=#{encoded_value}&"
+            buffer << encoded_key << "%5B%5D=" << encoded_value << "&"
```
philihp: Interpolating a string every iteration defeats the purpose of the buffer. 😄

PR #294, opened 2018-03-21
the method for redefinition top-level domain tld=. fixes #289

### LINKED ISSUE #289: the method for redefinition top-level domain `tld=`
it will be very convenient to use `tld=` to redefine the top-level domain, similar to redefine url path

```
require "addressable/uri"
=> true

url = Addressable::URI.parse("http://example.com/path/to/resource/")

# Example for path (method path=)

url.path
=> "/path/to/resource/"

url.to_s
=> "http://example.com/path/to/resource/"

url.path="redefenitioned/path/to/some/resource"
=> "redefenitioned/path/to/some/resource"

url.to_s
=> "http://example.com/redefenitioned/path/to/some/resource"

# Example for tld (expected method tld=)
irb(main):005:0> url.tld
=> "com"

irb(main):012:0> url.tld="uk"
NoMethodError: undefined method `tld=' for #<Addressable::URI:0x1e68a90>
Did you mean?  tld
        from (irb):12
        from C:/Ruby23/bin/irb.cmd:19:in `<main>'

# Got error, but expected:
=> "uk"
url.to_s
=> "http://example.uk/redefenitioned/path/to/some/resource"
```

Please, do the same method, for tld (tld=), it's convenient. Thank you

Mifrill:


spec/addressable/uri_spec.rb
```diff
@@ -5490,6 +5490,31 @@ def to_s
...
 
+describe Addressable::URI, "when given the tld " do
+  it "'uk' should have a tld of 'uk'" do
+    uri = Addressable::URI.parse("http://example.com")
+    uri.tld = "uk"
+
+    expect(uri.tld).to eq("uk")
+  end
+
+  context "witch " do
```
sporkmonger: Presumably you meant "which"?

Mifrill: Yup. Right, `which`, sorry.

`when given tld, which contains a subdomain`

I can by analogy make the method `url.basename=` and in the new PR correct this word, if you do not mind.

Thank you very much for the merge

PR #302, opened 2018-06-10
Be able to use #request_uri= without query

### LINKED ISSUE #300: unexpected zeroing of the path when using request_uri without query values
Thanks again for the excellent gem

The point is:
```
[1] pry(main)> require "addressable"
=> true
[2] pry(main)> url = Addressable::URI.parse("https://www.example.com")
=> #<Addressable::URI:0x2afce4b35158 URI:https://www.example.com>
[3] pry(main)> url.request_uri = "somepath/segment1/segment2"

=> "somepath/segment1/segment2"

[4] pry(main)> url.to_s
=> "https://www.example.com/"
```
I think, it would be logical to expect an error with the message that the string does not contain a query, or just keep the url is unchanged

or even better, in this case, do the same behavior as the method `path=`, expect:
```
url.request_uri = "somepath/segment1/segment2"
=> "somepath/segment1/segment2"
url.to_s
=> "https://www.example.com/somepath/segment1/segment2"
```
I would be grateful for the views on this matter, thanks

dentarg:
Close #300

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -1762,7 +1762,7 @@ module Addressable
           "Cannot set an HTTP request URI for a non-HTTP URI."
       end
       new_request_uri = new_request_uri.to_str
-      path_component = new_request_uri[/^([^\?]*)\?(?:.*)$/, 1]
+      path_component = new_request_uri[/^([^\?]*)\??(?:.*)$/, 1]
       query_component = new_request_uri[/^(?:[^\?]*)\?(.*)$/, 1]
       path_component = path_component.to_s
       path_component = (!path_component.empty? ? path_component : SLASH)
--- a/spec/addressable/uri_spec.rb
+++ b/spec/addressable/uri_spec.rb
@@ -1987,6 +1987,20 @@ describe Addressable::URI, "when parsed from " +
   end
 
   it "should have the correct request URI after assignment" do
+    @uri.request_uri = "/pub/WWW/TheProject.html?"
+    expect(@uri.request_uri).to eq("/pub/WWW/TheProject.html?")
+    expect(@uri.path).to eq("/pub/WWW/TheProject.html")
+    expect(@uri.query).to eq("")
+  end
+
+  it "should have the correct request URI after assignment" do
+    @uri.request_uri = "/some/where/else.html"
+    expect(@uri.request_uri).to eq("/some/where/else.html")
+    expect(@uri.path).to eq("/some/where/else.html")
+    expect(@uri.query).to eq(nil)
+  end
+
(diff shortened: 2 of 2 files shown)


Mifrill:
@dentarg  thank you, it so much better

sharshenov:
@sporkmonger Sorry for tagging you. But this PR fixes extremely annoying bug. Is there a chance this PR to be merged?

dentarg:
@sporkmonger [commented on the commit](https://github.com/sporkmonger/addressable/commit/4ccac35cdcc85303fb0378b60bf4b963012a5f28#commitcomment-29924127) and that got lost when I forced pushed to my branch.

> Probably would benefit from a test for setting query string to `“”`.

I didn't read this carefully enough and interpreted it as "a test for setting request uri to empty string", which was what I pushed:

```ruby
it "should have the correct request URI after assignment" do
  @uri.request_uri = ""
  expect(@uri.request_uri).to eq("/")
  expect(@uri.path).to eq("/")
  expect(@uri.query).to eq(nil)
end
```

I have now changed it to (which can also be seen in the diff here)

```ruby
it "should have the correct request URI after assignment" do
  @uri.query = ""
  expect(@uri.request_uri).to eq("/pub/WWW/TheProject.html?")
  expect(@uri.path).to eq("/pub/WWW/TheProject.html")
  expect(@uri.query).to eq("")
end
```

Is that what you thought of @sporkmonger?

sporkmonger:
Close. That's a worthwhile test I think, and maybe keep it, but I was actually looking for:

```ruby
@uri.request_uri = "/pub/WWW/TheProject.html?"
expect(@uri.query).to eq("")
```

dentarg:
Ah, I see. Fixed.

I didn't keep it, as I think it is covered by other specs, for example:

https://github.com/sporkmonger/addressable/blob/3ef784c4957b3ac50dcf3d8f7b9c957ee2e3e882/spec/addressable/uri_spec.rb#L2222-L2225

dentarg:
Re: coverage/coveralls check, is it confused by force pushes?

dentarg:
The build failure isn't related to this PR, only `gemfiles/Gemfile.public_suffix_2` that failed (https://travis-ci.org/sporkmonger/addressable/jobs/412992437), can't reproduce it locally, I guess some temporary issue?

dentarg:
> The build failure isn't related to this PR, only gemfiles/Gemfile.public_suffix_2 that failed (https://travis-ci.org/sporkmonger/addressable/jobs/412992437), can't reproduce it locally, I guess some temporary issue?

Yes it was

> Re: coverage/coveralls check, is it confused by force pushes?

My branch was behind master, I guess that was the reason. Branch rebased and all is green now.

PR #303, opened 2018-06-21
fixes #263 Be able to use #heuristic_parse for mailto URLs

### LINKED ISSUE #263: Addressable::URI#heuristic_parse breaks for mailto URLs
Hi,

I'm using `Addressable::URI` to parse the href target of `<a>` in arbitrary HTML documents. Sometimes the href is a mailto link. I'm running into an issue parsing these kinds of links.

Parse works for mailto URLs:
```ruby
uri = Addressable::URI.parse("mailto:test@test.com")
uri.to_hash
=> {:scheme=>"mailto", :user=>nil, :password=>nil, :host=>nil, :port=>nil, :path=>"test@test.com", :query=>nil, :fragment=>nil}
```

Heuristic parse does not:
```ruby
Addressable::URI.heuristic_parse("mailto:test@test.com")

Addressable::URI::InvalidURIError: Invalid character in host: 'test@test.com'
addressable-2.5.1/lib/addressable/uri.rb:2438:in `validate'
addressable-2.5.1/lib/addressable/uri.rb:2357:in `defer_validation'
addressable-2.5.1/lib/addressable/uri.rb:214:in `heuristic_parse'
```

I'm using heuristic parse because sometimes the links are incomplete, so I need the missing parts filled in.

Mifrill:


Mifrill:
fixes #263

sporkmonger:
Going to have to give a 👎 on this specific implementation. My general intent for `heuristic_parse` has been that it should largely behave the way a browser address bar's parser would. If you typed that email address into Chrome's address bar, you'd get a banner notice asking you if you meant to go to `http://user@domain.com/`, while opening a Google search for that value. It's certainly not 100% clear what the user intent is, but I'm quite wary about assuming an `@` means it's an email address in the absence of a scheme.

The other important property of `heuristic_parse` is that when intent is unambiguous, `heuristic_parse` should do the exact same thing as `parse`. This was actually the main reason I marked the original issue as accepted, because in your original example, intent was unambiguous because the scheme was provided, but it fails.

Consider the following URI, which is completely unambiguous (RFC 7565):

`acct:user@domain.com`

I think this PR might result in this being converted to a `mailto:` URI, which would be undesirable.

If you update this PR so that:

- `user@domain.com` becomes `mailto:user@domain.com`
- `mailto:user@domain.com` remains `mailto:user@domain.com`
- `acct:user@domain.com` remains `acct:user@domain.com`

And ensure there are tests for all 3 scenarios, then I can accept it. I'm a little on the fence about the first scenario, but since you can always compare the results of `heuristic_parse` with `parse`, I'm open to a more opinionated solution.

Mifrill:
@sporkmonger  Thank you very much for the detailed review/answer, I really appreciate it. And i'm sorry for delay, I was really obsessed with my job.

As far as I know, the browser should ignore everything that is specified before the @ symbol.
is obtained by entering the address, we logically should gather everything after the @ for `heuristic_parse`, coz your general intent, so I do not quite understand why:
> If you typed that email address into Chrome's address bar, you'd get a banner notice asking you if you meant to go to 

anyway, I did a test that reflects your thesis visually: "should ignore '@' if some scheme is present", if I understood it correctly.

Also i updated the PR according to your conditions. Many thanks

sporkmonger:
> As far as I know, the browser should ignore everything that is specified before the @ symbol.

The @ symbol in an HTTP context indicates the username for HTTP basic authentication. The browser does not ignore it.

spec/addressable/uri_spec.rb
```diff
@@ -6338,6 +6338,35 @@ def to_str
   end
 end
 
+describe Addressable::URI, "when given the input: 'user@domain.com'" do
+  before do
+    @input = "user@domain.com"
+  end
+
+  context "for heuristic parse" do
+    it "should remains 'mailto:user@domain.com'" do
```
sporkmonger: typo: s/remains/remain/

x2

spec/addressable/uri_spec.rb
```diff
@@ -6338,6 +6338,35 @@ def to_str
...
+      expect(uri.to_s).to   eq("mailto:user@domain.com")
+      expect(uri.scheme).to eq("mailto")
+    end
+
+    it "should remains 'acct:user@domain.com'" do
+      uri = Addressable::URI.heuristic_parse("acct:#{@input}")
+      expect(uri.to_s).to eq("acct:user@domain.com")
+    end
+
+    it "should ignore '@' if some scheme is present" do
```
sporkmonger: Test is good, but test description is wrong. See discussion comment.

spec/addressable/uri_spec.rb
```diff
@@ -6338,6 +6338,35 @@ def to_str
...
+      uri = Addressable::URI.heuristic_parse("http://#{@input}/")
+      expect(uri.to_s).to eq("http://user@domain.com/")
```
sporkmonger: Add: `expect(uri.user).to eq("user")`

spec/addressable/uri_spec.rb
Mifrill: ok, cleared it for all cases, thank you

spec/addressable/uri_spec.rb
Mifrill: Many thanks for the clarification. I understood what you mean

spec/addressable/uri_spec.rb
Mifrill: I did two separate tests for the context of HTTP, I thought it would be better. Thank you

toao:
any updates?

Mifrill:
@toao sorry for the delay. Thanks for reminding.

@sporkmonger I made the changes according to your previous review, please take look again. Thanks in advance.

PR #304, opened 2018-07-08
the method for redefinition basename

Mifrill:
Hi, thanks again for the gem

The point is: 
there are often situations where we need to change the basename

for example iteration through sections of the site:
```
url = Addressable::URI.parse("http://example.com/sections/categories/sub_categories/")

xpath("//a[@class='sub-categories']").each do |a|
  # when links contains only basenames
  a[:href] = "subcategory_1"
  # and also correct link should be "http://example.com/sections/categories/subcategory_1"
  # so in this case we have problems with fast and easy address replacement
  # especially if we need to save the `url` Addressable object, for later use in other methods

  # or we need to quickly set the basename and save it in the object, without re-parse
  # this is just few examples of which can be solved differently, but it seems to me, it would be more  convenient to have such a method directly here

  # example with method 'basename=':
  url.basename = a[:href]
  page.visit(url.to_s) # http://example.com/sections/categories/subcategory_1
end
```

I would be grateful for the views on this matter, thanks

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -1605,6 +1605,20 @@ module Addressable
     end
 
     ##
+    # Sets the basename component for this URI.
+    #
+    # @param [String, #to_str] new_basename. The new basename value
+    def basename=(new_basename)
+      self.path = begin
+        if path.empty? || path == "/"
+          new_basename
+        else
+          path.sub(%r{\b#{basename}(\/)?\z}, new_basename)
+        end
+      end
+    end
+
+    ##
     # The extname, if any, of the file in the path component.
     # Empty string if there is no extension.
     #
--- a/spec/addressable/uri_spec.rb
+++ b/spec/addressable/uri_spec.rb
@@ -5601,6 +5601,22 @@ describe Addressable::URI, "when given the tld " do
   end
 end
 
+describe Addressable::URI, "when given the input" \
+    "'http://test.com/category/sub_category/categories'" do
+  before do
+    @input = "http://test.com/category/sub_category/categories"
+  end
+
+  it "should have former path but with the new basename: 'category' "\
+     "when the replaceable basename is contained in uri path" do
+    # to prevent regressions of path, for example shouldn't be: http://test.com/category
+    uri = Addressable::URI.parse(@input)
+    uri.basename = "category"
+
+    expect(uri.to_s).to eq("http://test.com/category/sub_category/category")
+  end
+end
+
 describe Addressable::URI, "when given the path " +
     "'c:\\windows\\My Documents 100%20\\foo.txt'" do
   before do
@@ -6062,6 +6078,13 @@ describe Addressable::URI, "when given the input " +
     @input = "http://example.com/"
   end
 
+  it "should set the new basename: 'index.html'" do
+    uri = Addressable::URI.parse(@input)
+
+    expect(uri.basename = "index.html").to eq("index.html")
+    expect(uri.to_s).to eq("http://example.com/index.html")
+  end
+
   it "should heuristically parse to 'http://example.com/'" do
     @uri = Addressable::URI.heuristic_parse(@input)
     expect(@uri.to_s).to eq("http://example.com/")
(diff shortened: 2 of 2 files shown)


spec/addressable/uri_spec.rb
```diff
@@ -5515,6 +5515,47 @@ def to_s
...
+describe Addressable::URI, "When change the basename " do
+  it "should have the new basename: 'item'" do
+    uri = Addressable::URI.parse("http://test.com/some/path/index.html")
+
+    expect(uri.basename = "item").to eq("item")
+    expect(uri.to_s).to eq("http://test.com/some/path/item")
+  end
+end
+
+describe Addressable::URI, "After assignment the basename " do
```
sporkmonger: This doesn't read well: "Addressable::URI, after assignment the basename, it should have the correct basename".

Also, these shouldn't be grouped in the first place. They should be three independent `describe` blocks.

spec/addressable/uri_spec.rb
```diff
@@ -5515,6 +5515,47 @@ def to_s
   end
 end
 
+describe Addressable::URI, "When set the basename " do
```
sporkmonger: I don't think that the description you have here really conveys the scenario you're testing. It's a good test case, but your description of the test context is too vague. It's also a test context that exists elsewhere in the test suite. In general you should append to an existing test context when it already exists. In this case, there are definitely already test contexts for e.g. `http://example.com` which you can add a new `it` block to.

spec/addressable/uri_spec.rb
```diff
@@ -5515,6 +5515,47 @@ def to_s
...
+    uri.basename = "index.html"
+
+    expect(uri.basename).to eq("index.html")
+    expect(uri.to_s).to eq("http://test.com/index.html")
+  end
+end
+
+describe Addressable::URI, "When change the basename " do
+  it "should have the new basename: 'item'" do
+    uri = Addressable::URI.parse("http://test.com/some/path/index.html")
```
sporkmonger: Same issue as above. This text context is really similar to others in the test suite. You should append to those `describe` blocks rather than starting a new one.

spec/addressable/uri_spec.rb
```diff
@@ -5515,6 +5515,47 @@ def to_s
...
+    expect(uri.basename = "item").to eq("item")
+    expect(uri.to_s).to eq("http://test.com/some/path/item")
+  end
+end
+
+describe Addressable::URI, "After assignment the basename " do
+  it "should have the correct basename: 'subcat_1' "\
+    "when the current basename at the end contains a slash" do
+
+    url = "http://test.com/sections/categories/sub_categories/"
```
sporkmonger: I believe there are already `describe` blocks w/ paths containing trailing slashes.

lib/addressable/uri.rb
```diff
@@ -1558,6 +1558,15 @@ def basename
       return File.basename(self.path).sub(/;[^\/]*$/, EMPTY_STR)
     end
 
+    ##
+    # Sets the basename component for this URI.
+    #
+    # @param [String, #to_str] new_basename. The new basename value
+    def basename=(new_basename)
+      self.path = path.sub(%r{#{basename}(\/)?\z}, new_basename)
```
sporkmonger: I'd kind of like this regexp to ensure that the `basename` occurs at a boundary. Theoretically the `File.basename` call already addresses this, but in many other scenarios where you interpolate into a regexp, I'd expect to find a bug, and the only reason this isn't one is because `File.basename` finds the boundary. Adding a `\b` makes it more obvious that there isn't a bug here.

spec/addressable/uri_spec.rb
```diff
@@ -5515,6 +5515,47 @@ def to_s
...
+    uri = Addressable::URI.parse(url)
+
+    expect(uri.basename = "subcat_1").to eq("subcat_1")
+    expect(uri.to_s).to eq("http://test.com/sections/categories/subcat_1")
+  end
+
+  it "should have former path with the new basename: 'category' "\
+     "when the current basename is part of uri path" do
```
sporkmonger: I think of all your tests, this is the only one that should actually be its own unique `describe` block. However, please match the style of setting up a `describe` block used elsewhere in the test suite, where the URI object is initialized outside the `it` block and set as the test subject.

spec/addressable/uri_spec.rb
sporkmonger: Also it's not a URL when it's a `String` and a URI once it's been parsed. Please avoid this variable naming scheme. If you have to do this, use `uri_string` and `uri`.

spec/addressable/uri_spec.rb
sporkmonger: It's probably also a good idea to give a bit more explanation in a comment about why this test exists. It's a good test, particularly as insurance against possible regressions, but it took me awhile to understand why it existed. A clearly written comment would really help here.

Mifrill: ok, i got it, i'll write a clarification, thank you

spec/addressable/uri_spec.rb
Mifrill: yes, you are right, I'm sorry this is my inattention, here was a myself checked, in particular this part of regexp: `(\/)?\z`
it's redundant, i'l remove it, thank you

Mifrill:
Hi, @sporkmonger any chance to have it merged?

PR #314, opened 2018-08-19
Handle nil params better when using ? operator

### LINKED ISSUE #307: Partially expanded template misses the question mark when expanding to nil
Continuing https://github.com/sporkmonger/addressable/issues/200

```ruby
require 'addressable'
Addressable::Template.new('http://example.com{?a,b}').partial_expand(b: nil).pattern
# => "http://example.com{&a}"
```

therabidbanana:
Since nil params disappear from the template, we need to make sure reordering doesn't move a nil to the front and cause invalid urls. 

This essentially refines the work started in 
https://github.com/sporkmonger/addressable/pull/202 to make sure blank values don't get to be assigned the initial `?`.

Fixes #307


therabidbanana:
@sporkmonger - not sure how critical the 0.004% drop in code coverage is (I think it's because I deleted tested lines). I guess I could try to add a few more specs if we want it all green?

sporkmonger:
That’s not a problem.

therabidbanana:
@sporkmonger - did you want me to go ahead and merge this - not sure what your release process is?

PR #317, opened 2018-10-09
cleanup: replaced '!= nil' into shorter '!!'

Mifrill:


dentarg:
I don't think `!!` is the right thing to do. See https://github.com/rubocop-hq/ruby-style-guide#no-bang-bang

ashmaroli:
In many of the scenarios changed in this PR, neither `!= nil` nor `!!` is necessary..
```diff
- if processor != nil && processor.respond_to?(:restore)
+ if processor && processor.respond_to?(:restore)
```
From Ruby 2.3 onwards, one can simply use the SafeNavigation operator (`&.`):
```diff
- if processor != nil && processor.respond_to?(:restore)
+ if processor&.respond_to?(:restore)
```

Mifrill:
Yes, I think that we can do quite a lot of potential simplifications in the code, it seems to me better to start with a common config, for example rubocop, which suggested by @ashmaroli and then gradually make changes, because now the houndci-bot found a lot of offences in the legacy code, apparently it was for this reason added by the author. let's try to do everything green

PR #318, opened 2018-10-09
refactor: singlton methods of URI into dedicate file

Mifrill:
preamble:
the current uri.rb contains 2510 lines, it's a inconvenient when debugging, so i think that would be better to dedicated all of self.methods into separate file.

i'm not sure that is it better implementation against extend of module for those methods, but it's lgtm.
Besides it's help avoid confusion about alias_methods `class << self` which are repeated in 3 different places and also order of methods, for example: `protected` method `normalize_path` which unsuccessfully located below `protected`

if you have a different opinion, please tell me about your view, maybe you prefer to leave everything in monolith or you have another idea about class division, I would handle it if you don't mind.

Thank you for your time


Mifrill:
@dentarg yeah, got it. thank you
ok, but what about the refactoring of the current code?
for example make a code more readable, add rubocop config, the allocate of repetitive logical constructs into classes or modules, etc.
I have time and handle it now, what are you think?

PR #323, opened 2018-10-17
refactor: add module Validator

Mifrill:


lib/addressable/validator.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module Addressable
+  ##
+  # Ensures that the URI is valid.
+  module Validator
+    METHODS_POOL = %w[
```
sporkmonger: I'm not sure `_POOL` is the right suffix to use for this identifier.

Mifrill: yeah agree, thank you

sporkmonger:
I'm not in a rush to add this as it would be a breaking change for the `validate` method.

ashmaroli:
At a glance, I see some areas that could be improved:
```ruby
    def unescaped_backslashes?
      unreserved = URI::CharacterClasses::UNRESERVED
      sub_delims = URI::CharacterClasses::SUB_DELIMS
      regexp     = Regexp.new("^[#{unreserved}#{sub_delims}:]*$")
      !host[/^\[(.*)\]$/, 1].nil? && host[/^\[(.*)\]$/, 1] !~ regexp
    end
```
In the above code, `#{unreserved}#{sub_delims}` is interpolating *constants*.
So, why not create a new constant for the regex instead of allocating a new regex on each call to the `:unescaped_backslashes?` method?

```ruby
    def relative_path?
      path[0..0] != URI::SLASH
    end

    def path_with_two_leading_slashes?
      path[0..1] == URI::SLASH + URI::SLASH
    end
```
In the above two methods, would using `String#start_with?` be equivalent? If yes, that'd be better since using `String#[range]` allocates memory for the substring (on every call to the parent methods)

Mifrill:
@ashmaroli thank you for improved mind, nice points!

@sporkmonger please take a look at recent commits:
https://github.com/sporkmonger/addressable/pull/323/files/68a3f70a923dc79530f01f923c90f36201d426dd..f544148246a33f627978a85b5c6bdedd498da71b

Thank you for you time!

ashmaroli:
@Mifrill Part of the reason why I suggested creating a new constant instead of a memoized regexp (currently method def `:unreserved_sub_delims_regexp`) is because the regex in the constant would be created just once.
With a memoized method in a prepended module, there's a chance that the memoization is repeated  for every instance of the prepended instance. (I've not tested it, so not sure of the actual outcome).

My hypothesis is that as a result of *prepend*, method `:unreserved_sub_delims_regexp` is effectively available as an *instance_method* for `Addressable::URI` and therefore for every instance (via `Addressable::URI.new`) a new `unreserved_sub_delims_regexp` is generated and cached.

Feel free to experiment and correct me if I'm wrong.

Mifrill:
:notebook: I did some split changes and removed meta implementation to pass profile:memory check

PR #332, opened 2018-12-11
Improve readability of `uri.rb` with Whitespace

ashmaroli:
Only whitespace adjustments

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -39,19 +39,19 @@ module Addressable
     # Container for the character classes specified in
     # <a href="http://www.ietf.org/rfc/rfc3986.txt">RFC 3986</a>.
     module CharacterClasses
-      ALPHA = "a-zA-Z"
-      DIGIT = "0-9"
+      ALPHA      = "a-zA-Z"
+      DIGIT      = "0-9"
       GEN_DELIMS = "\\:\\/\\?\\#\\[\\]\\@"
       SUB_DELIMS = "\\!\\$\\&\\'\\(\\)\\*\\+\\,\\;\\="
-      RESERVED = GEN_DELIMS + SUB_DELIMS
+      RESERVED   = GEN_DELIMS + SUB_DELIMS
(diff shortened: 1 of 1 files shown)


dentarg:
I'm not sure about this. [My comment from #318](https://github.com/sporkmonger/addressable/pull/318#issuecomment-428338220):

> Another reason I'm hesitant about moving code around just for the sake of it, is that it makes the git history harder to work with. I think that it is especially important in a project like Addressable. It is old (long history) and it doesn't change often. I don't think we want to make working on Addressable even harder than it is.

(Handy link to diff without whitespace changes: https://github.com/sporkmonger/addressable/pull/332/files?w=1)

ashmaroli:
@dentarg Thank you for your feedback.
I'm aware of both the handy link to diff w/o whitespace changes and the general reluctance (in many projects on GitHub) to merge *pure whitespace* changes for the sake of Git history.

That's the reason I put the *focus on readability* here. There are still heavily crowded areas remaining but that will be for a future date e.g.:
https://github.com/sporkmonger/addressable/blob/52b7afc6262ec76030ba09ea0f59a71b269a77df/lib/addressable/uri.rb#L668-L683

sporkmonger:
I don't have an opinion. Addressable's overall style was originally established before the Ruby community had really settled in on strong opinions about consistent coding style, or whitespace, etc. It could use an update to whatever is current community preference, especially as Ruby versions have changed. But as I spend my day entirely in Go nowadays, it's something I'm largely indifferent to. I tend to defer to @dentarg here as he's massively more active on the project than I am now. Patrik, your call.

ashmaroli:
@dentarg I'll resolve conflicts if you plan to merge this in the near future. Else you may close.

dentarg:
While a might be a bit more active than @sporkmonger from time to time, because I do find addressable an interesting piece of software (and we use it at work too), I'm not able to dedicate very much time to it. So for the reasons I stated above, I think it is better to be passive with changes.

PR #335, opened 2019-02-08
Use Module#private_class_method instead of (class <<self; private SYMBOL; end)

ashmaroli:
Because it is simpler and intuitive.


luke-hill:
This #312 and #310 should all be merged in pronto, as they're all small PR's and have very easy to distinguish benefits (Speed or legibility).

cc/ @sporkmonger

Issue #254, opened 2017-01-26
Constants in Addressable::URI sometimes get mutated


QuinnWilton:
This is an issue I ran into when upgrading to 2.5, but I think other versions are affected in similar ways. Sometimes, when parsing a URL, some of the components of the parsed URL will be set directly to one of the constants defined in `Addressable::URI`. For example, if the path is "/", the instance variable will be set to `Addressable::URI::SLASH`.

If the parsed URL is then normalized, the encoding of `@normalized_path` is forced to be `Encoding::UTF_8`. This mutates the constant. This snippet demonstrates the behavior:

```
[1] pry(main)> Addressable::URI.parse("http://google.com").normalize
=> #<Addressable::URI:0x3fd9d0f0fe44 URI:http://google.com/>
[2] pry(main)> Addressable::URI::SLASH.freeze
=> "/"
[3] pry(main)> Addressable::URI.parse("http://google.com").normalize
RuntimeError: can't modify frozen String
```

The changed behavior is from this commit: https://github.com/sporkmonger/addressable/commit/8f4fee07243af0b788c625e59a9ca529489336d6

This is currently blocking my team from upgrading to 2.5, because we've found value in using the `IceNine` gem to `deep_freeze!` our parsed URLs from Addressable: on more than one occasion it has caught other dependencies of ours mutating URLs in unexpected ways. I understand that this is a non-standard usage of your gem, and understand if as a result, you do not consider this to be a bug.

In the meantime, we will most likely continue using 2.4, but it might be worth thinking about possible solutions here, if only because mutating constants is unexpected.

I think a possible fix would be to either check if the components are already UTF-8 encoded before calling `force_encoding` on them, or to move the `force_encoding` call into the `||= begin` block at the start of each of the `normalized_*` methods.

dentarg:
If not before, this has to be fixed to be compatible with [Ruby 3](https://bugs.ruby-lang.org/issues/11473). :)

```ruby
$ RUBYOPT="-w --enable-frozen-string-literal" irb
irb(main):001:0> require 'addressable/uri'
=> true
irb(main):002:0> Addressable::URI.parse("http://google.com").normalize
RuntimeError: can't modify frozen String
	from /Users/dentarg/.gem/ruby/2.3.3/gems/addressable-2.5.0/lib/addressable/uri.rb:439:in `force_encoding'
[...]

PR #340, opened 2019-04-18
Check if calling force_encoding is necessary before doing so in Addressable::URI, replace with encode! where applicable

ACBullen:
The first commit is a quick fix to [this issue](https://github.com/sporkmonger/addressable/issues/254), just adding in an extra check to avoid mutating strings when it isn't necessary. The second sets up handling for the off chance that the normalization process doesn't succeed at removing all of the utf-8 invalid characters. `force_encoding` just changes what the string reports its encoding is, using `str.encode!(Encoding::UTF_8, invalid: :replace, replace: "")` will change the coding and replace any invalid bytes with an empty string. It's possibly not strictly necessary, but it does seem to be the intent of forcing the encoding throughout this file.

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -552,7 +552,7 @@ module Addressable
       rescue ArgumentError
         encoded = self.encode_component(unencoded)
       end
-      encoded.force_encoding(Encoding::UTF_8)
+      encoded.encode!(Encoding::UTF_8, invalid: :replace, replace: "")
       return encoded
     end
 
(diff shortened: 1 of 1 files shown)


dentarg:
@ACBullen do you mind removing the "pullbot" commits from the history here? I.e. Just keeping bfc9a448b41195401045d5db21461391d276da2a 686dbbbfc7a9ec67e40826c8bde8661e8e6a7471. I think that's make sense if this is to be merged.

ACBullen:
@dentarg Removed those commits from history

ACBullen:
It looks like the second commit is causing compatibility issues with previous ruby versions so I am going to reopen this PR with just that first commit in a separate branch.

PR #341, opened 2019-04-23
Check Encodings before calling force_encoding in Addressable::URI

### LINKED ISSUE #254: Constants in Addressable::URI sometimes get mutated

ACBullen:
Reopening [this pr](https://github.com/sporkmonger/addressable/pull/340) with just the minimum changes to fix the issue blocking our use of the current version.

This is a quick fix to [this issue](https://github.com/sporkmonger/addressable/issues/254), just adding in an extra check to avoid mutating strings when it isn't necessary

Close #254

### DIFF
--- a/spec/addressable/uri_spec.rb
+++ b/spec/addressable/uri_spec.rb
@@ -1060,9 +1060,7 @@ describe Addressable::URI, "when normalized and then deeply frozen" do
   end
 
   it "should not allow destructive operations" do
-    ruby_check = Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.5.0")
-    error = ruby_check ? RuntimeError : FrozenError
-    expect { @uri.normalize! }.to raise_error(error)
+    expect { @uri.normalize! }.to raise_error(RuntimeError)
   end
 end
 
--- a/spec/addressable/uri_spec.rb
+++ b/spec/addressable/uri_spec.rb
@@ -1061,7 +1061,9 @@ describe Addressable::URI, "when normalized and then deeply frozen" do
   end
 
   it "should not allow destructive operations" do
-    expect { @uri.normalize! }.to raise_error(FrozenError)
+    ruby_check = Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.5.0")
+    error = ruby_check ? RuntimeError : FrozenError
+    expect { @uri.normalize! }.to raise_error(error)
   end
 end
 
--- a/spec/addressable/uri_spec.rb
+++ b/spec/addressable/uri_spec.rb
@@ -999,6 +999,72 @@ describe Addressable::URI, "when frozen" do
   end
 end
 
+describe Addressable::URI, "when normalized and then deeply frozen" do
+  before do
+    @uri = Addressable::URI.parse(
+      "http://user:password@example.com:8080/path?query=value#fragment"
+    ).normalize!
+
+    @uri.instance_variables.each do |var|
+      @uri.instance_variable_set(var, @uri.instance_variable_get(var).freeze)
+    end
+
+    @uri.freeze
+  end
+
+  it "#normalized_scheme should not error" do
+    expect { @uri.normalized_scheme }.not_to raise_error
+  end
+
+  it "#normalized_user should not error" do
+    expect { @uri.normalized_user }.not_to raise_error
+  end
+
+  it "#normalized_password should not error" do
+    expect { @uri.normalized_password }.not_to raise_error
+  end
+
+  it "#normalized_userinfo should not error" do
+    expect { @uri.normalized_userinfo }.not_to raise_error
+  end
+
+  it "#normalized_host should not error" do
+    expect { @uri.normalized_host }.not_to raise_error
+  end
+
+  it "#normalized_authority should not error" do
+    expect { @uri.normalized_authority }.not_to raise_error
+  end
+
+  it "#normalized_port should not error" do
+    expect { @uri.normalized_port }.not_to raise_error
+  end
+
+  it "#normalized_site should not error" do
+    expect { @uri.normalized_site }.not_to raise_error
+  end
+
+  it "#normalized_path should not error" do
+    expect { @uri.normalized_path }.not_to raise_error
+  end
+
+  it "#normalized_query should not error" do
+    expect { @uri.normalized_query }.not_to raise_error
+  end
+
+  it "#normalized_fragment should not error" do
+    expect { @uri.normalized_fragment }.not_to raise_error
+  end
+
+  it "should be frozen" do
+    expect(@uri).to be_frozen
+  end
+
+  it "should not allow destructive operations" do
+    expect { @uri.normalize! }.to raise_error(FrozenError)
+  end
+end
+
 describe Addressable::URI, "when created from string components" do
   before do
     @uri = Addressable::URI.new(
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -900,7 +900,9 @@ module Addressable
         end
       end
       # All normalized values should be UTF-8
-      @normalized_scheme.force_encoding(Encoding::UTF_8) if @normalized_scheme
+      if @normalized_scheme && @normalized_scheme.encoding != Encoding::UTF_8
+        @normalized_scheme.force_encoding(Encoding::UTF_8)
+      end
       @normalized_scheme
     end
 
@@ -955,7 +957,9 @@ module Addressable
         end
       end
       # All normalized values should be UTF-8
-      @normalized_user.force_encoding(Encoding::UTF_8) if @normalized_user
+      if @normalized_user && @normalized_user.encoding != Encoding::UTF_8
+        @normalized_user.force_encoding(Encoding::UTF_8)
+      end
       @normalized_user
     end
 
@@ -1012,7 +1016,9 @@ module Addressable
         end
       end
       # All normalized values should be UTF-8
-      if @normalized_password
+      if @normalized_password && (
+        @normalized_password.encoding != Encoding::UTF_8
+      )
         @normalized_password.force_encoding(Encoding::UTF_8)
       end
       @normalized_password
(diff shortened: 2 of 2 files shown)


dentarg:
> Reopening [this pr](https://github.com/sporkmonger/addressable/pull/340) with just the minimum changes to fix the issue blocking our use of the current version.

I'm curious, what are the issue you are facing with the current version?

> This is a quick fix to [this issue](https://github.com/sporkmonger/addressable/issues/254), just adding in an extra check to avoid mutating strings when it isn't necessary

https://github.com/sporkmonger/addressable/issues/254 is about constants, but this PR only touches instance variables, so I don't think it is the same.

ACBullen:
The issue with constants is sort of an extension of the issue we have right now since we are normalizing the `Addressable::URI` object and then deep freezing. Even if the path or other part of the uri doesn't map to a constant, we are marking all of those strings as intended to be immutable. Since the `force_encoding` runs every time we try to reference parts of the parsed and normalized URI, this causes the same `RuntimeError: can't modify frozen String` to return. 

Without this check, after the strings go through normalization, they are in their final state and do not mutate unless something tries to reassign one of the.

dentarg:
Got it, thanks.

I think we need to add tests for this.

baseballlover723:
@dentarg I have rebased and updated this PR. Let me know if theres anything else you want me to change.

spec/addressable/uri_spec.rb
```diff
@@ -999,6 +999,72 @@ def to_s
...
+  it "#normalized_fragment should not error" do
+    expect { @uri.normalized_fragment }.not_to raise_error
+  end
+
+  it "should be frozen" do
+    expect(@uri).to be_frozen
+  end
+
+  it "should not allow destructive operations" do
+    expect { @uri.normalize! }.to raise_error(FrozenError)
```
dentarg, line 1064: `FrozenError` only exist [in Ruby 2.5 and newer](https://docs.ruby-lang.org/en/2.5.0/NEWS.html) (I just learned this too). Sounds like it was `RuntimeError` before: https://github.com/jruby/jruby/pull/5106

Hence the test failures. Do you mind addressing this?

baseballlover723: I just pushed some changes

baseballlover723:
@dentarg This is ready for review again

spec/addressable/uri_spec.rb
```diff
@@ -998,6 +998,74 @@ def to_s
...
+    expect { @uri.normalized_fragment }.not_to raise_error
+  end
+
+  it "should be frozen" do
+    expect(@uri).to be_frozen
+  end
+
+  it "should not allow destructive operations" do
+    ruby_check = Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.5.0")
+    error = ruby_check ? RuntimeError : FrozenError
```
sporkmonger, line 1064: I'd prefer to have this test not care at all which error is raised.

sporkmonger (changes requested):

spec/addressable/uri_spec.rb
dentarg: Not sure I'm following... it says `should not allow destructive operations` and we expect `@uri.normalize!` to raise RuntimeError/FrozenError (`can't modify frozen String`). Is that because you shouldn't be able to use `#normalize!` on a frozen `Addressable::URI`? Maybe we can make that even more clear in the test? (The `@uri.freeze` happens way up in the `before` block so it is easy to miss)

Maybe `Addressable::URI#normalize!` should detect that the URI is frozen and raise an error defined by Addressable? (`FrozenURIError` perhaps?)

---

Ah, this particular spec doesn't test the changes made in this branch, I ran this against Addressable 2.8.0

```ruby
# Ruby 2.4.3

irb(main):005:0* @uri = Addressable::URI.parse("http://user:password@example.com:8080/path?query=value#fragment")
=> #<Addressable::URI:0x3ff6e204dd78 URI:http://user:password@example.com:8080/path?query=value#fragment>
irb(main):006:0> @uri.freeze
=> #<Addressable::URI:0x3ff6e204dd78 URI:http://user:password@example.com:8080/path?query=value#fragment>
irb(main):007:0> @uri.normalize!
RuntimeError: can't modify frozen Addressable::URI
	from /Users/dentarg/.gem/ruby/2.4.3/gems/addressable-2.8.0/lib/addressable/uri.rb:2518:in `remove_instance_variable'
	from /Users/dentarg/.gem/ruby/2.4.3/gems/addressable-2.8.0/lib/addressable/uri.rb:2518:in `block in replace_self'
	from /Users/dentarg/.gem/ruby/2.4.3/gems/addressable-2.8.0/lib/addressable/uri.rb:2516:in `each'
	from /Users/dentarg/.gem/ruby/2.4.3/gems/addressable-2.8.0/lib/addressable/uri.rb:2516:in `replace_self'
	from /Users/dentarg/.gem/ruby/2.4.3/gems/addressable-2.8.0/lib/addressable/uri.rb:2211:in `normalize!'
	from (irb):7
	from /Users/dentarg/.rubies/2.4.3/bin/irb:11:in `<main>'
```

```ruby
# Ruby 3.1.2

irb(main):005:0> @uri = Addressable::URI.parse("http://user:password@example.com:8080/path?query=value#fragment")
=> #<Addressable::URI:0xb43c URI:http://user:password@example.com:8080/path?query=value#fragment>
irb(main):006:0>
irb(main):007:0> @uri.freeze
=> #<Addressable::URI:0xb43c URI:http://user:password@example.com:8080/path?query=value#fragment>
irb(main):008:0> @uri.normalize!
/Users/dentarg/.gem/ruby/3.1.2/gems/addressable-2.8.0/lib/addressable/uri.rb:2518:in `remove_instance_variable': can't modify frozen Addressable::URI: #<Addressable::URI:0xb43c URI:http://user:password@example.com:8080/path?query=value#fragment> (FrozenError)
	from /Users/dentarg/.gem/ruby/3.1.2/gems/addressable-2.8.0/lib/addressable/uri.rb:2518:in `block in replace_self'
	from /Users/dentarg/.gem/ruby/3.1.2/gems/addressable-2.8.0/lib/addressable/uri.rb:2516:in `each'
	from /Users/dentarg/.gem/ruby/3.1.2/gems/addressable-2.8.0/lib/addressable/uri.rb:2516:in `replace_self'
	from /Users/dentarg/.gem/ruby/3.1.2/gems/addressable-2.8.0/lib/addressable/uri.rb:2211:in `normalize!'
	from (irb):8:in `<main>'
	from /Users/dentarg/.rubies/3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
	from /Users/dentarg/.rubies/3.1.2/bin/irb:25:in `load'
	from /Users/dentarg/.rubies/3.1.2/bin/irb:25:in `<main>'
```

Perhaps we shouldn't change this behaviour as I suggested above

---

@sporkmonger should we just check error message? Expect `can't modify frozen Addressable::URI`. Those can change from Ruby to Ruby version too, but probably unlikely in this case IMHO.

sporkmonger: Mostly I'd prefer to have the test make the same checks regardless of Ruby version. If the error message is consistent between versions but the type is not, it seems fine to check the error message, but if that's inconsistent too, I think it's fine to verify that an error, any error, has been thrown.

dentarg: We can have the test make the same checks regardless of Ruby version. If we look at the hierarchy of exceptions at https://ruby-doc.org/core-2.6.5/Exception.html we can see that FrozenError is a subclass of RuntimeError. That should never change really. We can use `#is_a?` and always check if it is a RuntimeError:

```ruby
irb(main):007:1* ex = begin
irb(main):008:1*   raise FrozenError, "can't modify frozen String"
irb(main):009:1* rescue => e
irb(main):010:1*   e
irb(main):011:0> end
=> #<FrozenError: can't modify frozen String>
irb(main):012:0> ex.inspect
=> "#<FrozenError: can't modify frozen String>"
irb(main):013:0> ex.is_a?(RuntimeError)
=> true
```

spec/addressable/uri_spec.rb
```diff
@@ -998,6 +998,75 @@ def to_s
...
+  end
+
+  it "should be frozen" do
+    expect(@uri).to be_frozen
+  end
+
+  it "should not allow destructive operations" do
+    expect { @uri.normalize! }.to raise_error { |error|
+      expect(error.message).to match(/can't modify frozen/)
+      expect(error).to satisfy { |e| RuntimeError === e || TypeError === e }
```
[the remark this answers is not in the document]
dentarg: See https://github.com/sporkmonger/addressable/pull/341#discussion_r934145690

spec/addressable/uri_spec.rb
dentarg: (We don't need to check the message)

baseballlover723: Are there any other changes required? I've implemented the above changes.

dentarg: Looks good now, thanks!

dentarg (approved):

PR #343, opened 2019-05-16
Add RubyGems.org-displayed metadata to gemspec

olleolleolle:
This PR adds [RubyGems metadata](https://guides.rubygems.org/specification-reference/#metadata) to make it possible to navigate from RubyGems _directly_ into the CHANGELOG document via the neat link they surface.

### DIFF
--- a/addressable.gemspec
+++ b/addressable.gemspec
@@ -34,4 +34,14 @@ Gem::Specification.new do |s|
     s.add_dependency(%q<public_suffix>, ["< 4.0", ">= 2.0.2"])
     s.add_dependency(%q<bundler>, ["< 3.0", ">= 1.0"])
   end
+
+  if s.respond_to?(:metadata)
+    s.metadata = {
+      "bug_tracker_uri" => "https://github.com/sporkmonger/addressable/issues",
+      "changelog_uri" => "https://github.com/sporkmonger/addressable/blob/master/CHANGELOG.md",
+      "documentation_uri" => "http://rubydoc.info/gems/addressable",
+      "homepage_uri" => "https://github.com/sporkmonger/addressable",
+      "source_code_uri" => "https://github.com/sporkmonger/addressable"
+    }
+  end
 end


dentarg:
Nice @olleolleolle 

Do you know if these values from the gemspec file will override whats currently configured over at https://rubygems.org/gems/addressable? I noticed that there are a few links there already, but they are non HTTPS (so I assume they were entered into the web interface a long time ago), would be nice if they were updated when a new version of the gem is published.

I see that RubyGems page for addressable links to http://rubydoc.info/gems/addressable for "Documentation". I think we can add that to the gemspec too. Do you mind adding https://www.rubydoc.info/gems/addressable as `documentation_uri`?

olleolleolle:
I'll add as many as I can find. 

The gemspec metadata is nowadays the interface to this. It makes sense to add all of it.

olleolleolle:
@dentarg Turns out there was no wiki content in this repo. Perhaps that can even be disabled?

dentarg:
Yeah, I saw that the wiki is empty here. I can't disabled it (or change anything on rubygems.org), but I will merge this so we can benefit from it with the next release. Thanks!

ashmaroli:
> Perhaps that can even be disabled?

@dentarg @olleolleolle probably meant disabling the wiki on this repository..

dentarg:
Bad phrasing by me but I got that and I can't do it

dentarg:
@olleolleolle sorry, I just realized that these changes should have been made to the rake task that generates the gemspec

https://github.com/sporkmonger/addressable/blob/9a02ca644ebe7c8a0a139380f3765dcd6db9edc6/tasks/gem.rake#L5-L10

(Because the wasn't there, these changes were removed in https://github.com/sporkmonger/addressable/commit/e087d7de14bebd6cb7c27e2a792bef767f1f118c#diff-76a9e5bbb0f907a004d60b3a55b1460a)

PR #344, opened 2019-05-23
Add URI.abolute_uri? to test if input is absolute?

ashmaroli:
To reduce memory-allocations and therefore speed  up the operation.

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -769,6 +769,22 @@ module Addressable
     end
 
     ##
+    # Determines if given string is an absolute URI.
+    #
+    # An optimized alternative to <code>URI.parse(str).absolute?</code> that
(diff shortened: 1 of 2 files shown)


ashmaroli:
@dentarg Any feedback on this..?

sporkmonger:
To be honest I'm not fully convinced. I'm all for performance optimization where it makes sense, but I'm pretty reluctant to change the API surface just to get a performance boost in a specific scenario.

ashmaroli:
We introduced a cache (and memoizations at some places) around the code employing `Addressable`.

So while I still feel it would be *better* for testing whether given `uri` is absolute or not by just checking if the `uri` has a valid `scheme` pattern`, instead of parsing it (which is what creates numerous intermediate objects) into an `Addressable::URI` object and then checking its attribute, this may no longer be needed for our use-cases.

PR #345, opened 2019-06-13
Add :compacted option to URI#normalized_query

pushrax:
This option removes empty parameters, having no name or value.

### DIFF
--- a/spec/addressable/uri_spec.rb
+++ b/spec/addressable/uri_spec.rb
@@ -4303,6 +4303,26 @@ describe Addressable::URI, "when parsed from " +
   end
 end
 
+describe Addressable::URI, "when parsed from 'http://example/?b=1&a=2&c=3'" do
+  before do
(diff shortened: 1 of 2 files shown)


spec/addressable/uri_spec.rb
```diff
@@ -4303,6 +4303,28 @@ def to_s
...
+describe Addressable::URI, "when parsed from " +
+    "'http://example.com/?b=1&a=2&c=3'" do
+  before do
+    @uri = Addressable::URI.parse("http://example.com/?b=1&a=2&c=3")
+  end
+
+  it "should have a sorted normalized query of 'a=2&b=1&c=3'" do
+    expect(@uri.normalized_query(:sorted)).to eq("a=2&b=1&c=3")
+  end
+end
```
pushrax: Seemed like there was no test coverage for `:sorted` so I added it while in here.

PR #349, opened 2019-07-03
Fix failing Travis builds

### LINKED ISSUE #348: Fix broken builds
As seen on https://travis-ci.org/sporkmonger/addressable/builds/551279504 the tests for old rubies fail, due to missing idn library, when `idn-ruby` is to be installed for these versions:

https://github.com/sporkmonger/addressable/blob/60e119e12624475870e23af1b79905b037ebc48f/Gemfile#L32

https://github.com/sporkmonger/addressable/blob/60e119e12624475870e23af1b79905b037ebc48f/gemfiles/Gemfile.public_suffix_2#L34

The builds on Travis CI [now run on xenial](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment) (Ubuntu 16.04) and not trusty (Ubuntu 14.04), and I guess that is the reason for the idn library missing.

The jruby failures seems to be https://github.com/rvm/rvm/issues/4720

ashmaroli:
Resolves #348

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,10 @@ matrix:
     - gemfile: gemfiles/Gemfile.public_suffix_2
       rvm: 2.5.5
   allow_failures:
+    - rvm: jruby-9.0.5.0
     - rvm: ruby-head
     - rvm: ruby-head-clang
   fast_finish: true
 before_install:
-  - gem update bundler
-#   - apt-get update
-#   - apt-get install idn
+  - sudo apt-get update
+  - sudo apt-get install libidn11-dev


.travis.yml
```diff
@@ -19,10 +19,16 @@ matrix:
...
 before_install:
-  - gem update bundler
-#   - apt-get update
-#   - apt-get install idn
+  - sudo apt-get update
+  - sudo apt-get install idn
+  - sudo apt-get install libldap2-dev
+  - sudo apt-get install libidn11-dev
+  - gem uninstall bundler -x --force || true
+  - gem install bundler --version 1.17.3
```
dentarg: Why do we need this version of bundler?

.travis.yml
```diff
@@ -19,10 +19,16 @@ matrix:
...
     - rvm: ruby-head-clang
   fast_finish: true
 before_install:
-  - gem update bundler
-#   - apt-get update
-#   - apt-get install idn
+  - sudo apt-get update
+  - sudo apt-get install idn
+  - sudo apt-get install libldap2-dev
+  - sudo apt-get install libidn11-dev
```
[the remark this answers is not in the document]
ashmaroli: I'm not entirely sure why package `libidn11-dev` works over package `idn`.

.travis.yml
```diff
@@ -19,10 +19,12 @@ matrix:
     - gemfile: gemfiles/Gemfile.public_suffix_2
       rvm: 2.5.5
   allow_failures:
+    - rvm: jruby-9.0.5.0
```
dcermak: This version of jruby doesn't seem to work with java 11, which is the default on Travis now. I'd therefore propose to drop it as it is officially unsupported.

.travis.yml
```diff
@@ -19,10 +19,12 @@ matrix:
...
+    - rvm: jruby-9.1.17.0
```
dcermak: This version of jruby doesn't appear to be causing any issues anymore. The same applies to `9.2.6.0`.

.travis.yml
ashmaroli: @dentarg Your call..

sporkmonger: I think we should just drop it. Allow failures makes sense if you expect it might sometimes pass. If you expect it to _always_ fail, it doesn't offer much value to run the test at all.

PR #350, opened 2019-07-10
de-dupe parameters when normalizing query

MothOnMars:
This PR is intended to de-duplicate parameters from urls such as: `&foo=bar&foo=bar`. It preserves parameters with the same key but differing values, such as `&foo=bar&foo=baz`.


sporkmonger:
Interesting proposal, but seems reasonable. This would be a breaking change for normalization, so I'd be interested to hear feedback from other stakeholders.

MothOnMars:
I saw that the `compacted` flag was just added. It might make sense to limit this to compacted normalization, i.e.:
```ruby
pairs.delete_if(&:empty?).uniq if flags.include?(:compacted)
```

dentarg:
> Interesting proposal, but seems reasonable. This would be a breaking change for normalization, so I'd be interested to hear feedback from other stakeholders.

Yeah, I was thinking the same. While it looks like a resonable thing to do, addressable has been around so long, so I think it is best not to change the normalization behavior.

> It might make sense to limit this to compacted normalization

Yeah, that could work I think. That feature hasn't been around long so adding this behavior to it should have much less impact.

MothOnMars:
@dentarg / @sporkmonger - I've updated this PR with a commit to ensure that the de-duping will only occur when the `:compacted` flag is in use:
https://github.com/sporkmonger/addressable/pull/350/commits/314df8fe01e6f9a41910eb58368e710dc878e3f8

PR #311, opened 2018-08-17
Use Rubocop to lint quote style for strings

ashmaroli:
Since `houndci-bot` prefers double-quoted strings, we can use Rubocop to use that style consistently across the codebase.

Invocations:
```bash
# Run Rubocop against this gem
bundle exec rubocop

# Automatically convert single-quoted strings to double-quoted strings
# (Every other "cop" has been disabled in `.rubocop_todo.yml`)
bundle exec rubocop --autocorrect
```

PR #354, opened 2019-08-31
Pulling rubocop config from #311.

sporkmonger:


sporkmonger:
This is more of a yak than I want to address right now.

PR #355, opened 2019-08-31
Auto-corrected style fixes

sporkmonger:


sporkmonger:
This is going to be a big yak.

PR #360, opened 2019-09-09
Test against TruffleRuby

rafaelfranca:
Test against the TruffleRuby implementation.

All tests are green already.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ language: ruby
 rvm:
   - jruby-9.1.17.0
   - jruby-9.2.6.0
+  - truffleruby
   - ruby-head
   - ruby-head-clang
   - 2.0.0


dentarg:
Sure, doesn't hurt I think... do you know why TruffleRuby is so slow?

rafaelfranca:
I'll still investigate why running the tests takes 2 minutes compared with the 19s of jruby, but part of why the build is slow are the 200s on bundle install that is fixed by the cache.

dentarg:
> 200s on bundle install

Looks like a known issue: https://github.com/oracle/truffleruby/issues/1398

dentarg:
And it looks like others have the same experience running their tests: https://github.com/oracle/truffleruby/issues/1415 (I added the difference for addressable there)

PR #361, opened 2019-09-10
Remove changelog reference to unmerged pull request

ashmaroli:
The entry refers to #323 which is not part of Addressable 2.7.0


dentarg:
Good catch

sporkmonger:
My bad!

PR #362, opened 2019-10-03
Pass a block to expect instead of a lambda

rafaelfranca:
Since https://github.com/rspec/rspec-expectations/commit/771da6d491cd4bf66c3e3d91e68f253b3408af41 rspec will raise if you pass a value trying to use the block expectation.

### DIFF
--- a/spec/addressable/uri_spec.rb
+++ b/spec/addressable/uri_spec.rb
@@ -65,116 +65,116 @@ end
 
 describe Addressable::URI, "when created with a non-numeric port number" do
   it "should raise an error" do
-    expect(lambda do
+    expect do
       Addressable::URI.new(:port => "bogus")
-    end).to raise_error(Addressable::URI::InvalidURIError)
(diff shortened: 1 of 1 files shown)


dentarg:
> Since [rspec/rspec-expectations@771da6d](https://github.com/rspec/rspec-expectations/commit/771da6d491cd4bf66c3e3d91e68f253b3408af41) rspec will raise if you pass a value trying to use the block expectation.

The change in rspec-expectations [has been reverted](https://github.com/rspec/rspec-expectations/pull/1125#issuecomment-538960475), it was [released in 3.8.5 but rolled back in 3.8.6](https://github.com/rspec/rspec-expectations/blob/74871dfec85e656cddc80672f8b1eaf766651cf6/Changelog.md#386--2019-10-07).

But the change here still might make sense?

rafaelfranca:
I think they do. At least we are saving a few characters 😄

pirj (dismissed):
👍

PR #368, opened 2019-11-26
Freeze string literals

PeterGillis:
Hi there Sporkmonger,

My Pull-request deletes the `.freeze` from the gemspec file and instead adds to the top of the file the magic comment:

`# frozen_string_literal: true`

This should solve the freeze issue and makes the file a little cleaner.

I hope that helps.

### DIFF
--- a/addressable.gemspec
+++ b/addressable.gemspec
@@ -13,13 +13,7 @@ Gem::Specification.new do |s|
   s.description = "Addressable is an alternative implementation to the URI implementation that is\npart of Ruby's standard library. It is flexible, offers heuristic parsing, and\nadditionally provides extensive support for IRIs and URI templates.\n"
   s.email = "bob@sporkmonger.com"
   s.extra_rdoc_files = ["README.md"]
-  s.files = [
-    "CHANGELOG.md", "Gemfile", "LICENSE.txt", "README.md", "Rakefile", "data/unicode.data", "lib/addressable", "lib/addressable.rb",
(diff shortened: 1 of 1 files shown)


dentarg:
> This should solve the freeze issue

What issue is that?

FYI, the gemspec is generated by rake, see the Rakefile and [its task](https://github.com/sporkmonger/addressable/tree/master/tasks)

PeterGillis:
Hi there, 
Sorry for not being clear, let me clarify. 
When I found the gem (it is a dependency of another gem I am using) I noticed that there was a large amount of `.freeze' statements.
I should not have said 'freeze issue' I just meant that maybe using `# frozen_string_literal: true` could avoid using `.freeze` so much. A small suggestion, not a solution to a problem.

dentarg:
@PeterGillis I see, are you able to provide steps to reproduce that problem?

Then we can try to understand if it is possible to fix in addressable

PeterGillis:
@dentarg, there's no problem, sorry I think you misunderstood. 
My PR is a proposal to re-factor, to make the code more DRY. 
Substituting writing `.freeze` multiple times and instead writing it in the magic comment once only.
Magic comment meaning this:
`# frozen_string_literal: true` . 
See [this:](https://stackoverflow.com/questions/37799296/what-does-the-comment-frozen-string-literal-true-do)
There's no 'problem' here to reproduce.

dentarg:
@PeterGillis ah got it, I miss understood 

But you can't make this change here, as the gemspec is generated by code, see my first comment

Issue #363, opened 2019-10-24
Hostname not supplied


noff:
Hello. Here is an example:

```
Addressable::URI.parse("https://user:password@dev.domain.com").omit(:scheme, :port, :user, :password, :host)
Traceback (most recent call last):
        1: from (irb):22
Addressable::URI::InvalidURIError (Hostname not supplied: '')
```

It raises only with `omit` method.

When try to parse URL and use an object, everything works great:

```
u = Addressable::URI.parse("https://user:password@dev.domain.com").
u.path
=> ""
```

noff:
```addressable (~> 2.3, >= 2.3.8)```

**Comment by noff** [comment]
Just updated to `addressable (2.7.0)`, same issue

**Comment by dentarg** [comment]
> `Addressable::URI.parse("https://user:password@dev.domain.com").omit(:scheme, :port, :user, :password, :host)`

@noff what do you expect this to return?

This?

```ruby
irb(main):009:0> Addressable::URI.parse("")
=> #<Addressable::URI:0x168 URI:>
```

---

Tested this myself with a few variants, seems to only trigger with username/password in the URL

```ruby
irb(main):016:0> Addressable::URI.parse("https://foo.com").omit(:scheme)
=> #<Addressable::URI:0x1e0 URI://foo.com>
irb(main):017:0> Addressable::URI.parse("https://foo.com").omit(:scheme, :host)
=> #<Addressable::URI:0x1f4 URI:>
irb(main):018:0> Addressable::URI.parse("https://foo.com").omit(:scheme, :port, :host)
=> #<Addressable::URI:0x208 URI:>
irb(main):019:0> Addressable::URI.parse("https://foo.com").omit(:scheme, :port, :user, :host)
=> #<Addressable::URI:0x21c URI:>
irb(main):020:0> Addressable::URI.parse("https://foo.com").omit(:scheme, :port, :user, :password, :host)
=> #<Addressable::URI:0x230 URI:>
irb(main):021:0> Addressable::URI.parse("https://foo@foo.com").omit(:scheme, :port, :user, :password, :host)
=> #<Addressable::URI:0x244 URI:>
irb(main):022:0> Addressable::URI.parse("https://foo:bar@foo.com").omit(:scheme, :port, :user, :password, :host)
Traceback (most recent call last):
        7: from /Users/dentarg/.rubies/ruby-2.7.3/bin/irb:23:in `<main>'
[...]

PR #369, opened 2019-12-03
Avoid host validate when is omitted

JuarezLustosa:
Fix issue #363

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -74,6 +74,9 @@ module Addressable
       "ldap" => 389,
       "prospero" => 1525
     }
+    ##
+    #  @@avoid_host_omitted is to avoid host validate when is set to be omitted
+    @@avoid_host_omitted = true
 
     ##
     # Returns a URI object based on the parsed string.
@@ -2292,11 +2295,15 @@ module Addressable
         :scheme, :user, :password, :userinfo, :host, :port, :authority,
         :path, :query, :fragment
       ]
+
       unless invalid_components.empty?
         raise ArgumentError,
           "Invalid component names: #{invalid_components.inspect}."
       end
       duplicated_uri = self.dup
+
+      @avoid_host_omitted = !components.include?(:host)
+
       duplicated_uri.defer_validation do
         components.each do |component|
           duplicated_uri.send((component.to_s + "=").to_sym, nil)
@@ -2449,10 +2456,8 @@ module Addressable
         raise InvalidURIError,
           "Absolute URI missing hierarchical segment: '#{self.to_s}'"
       end
-      if self.host == nil
-        if self.port != nil ||
-            self.user != nil ||
-            self.password != nil
+      if host.nil? && @@avoid_host_omitted
+        if !port.nil? || !user.nil? || !password.nil?
           raise InvalidURIError, "Hostname not supplied: '#{self.to_s}'"
         end
       end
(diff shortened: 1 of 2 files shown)


lib/addressable/uri.rb
```diff
@@ -2403,6 +2410,7 @@ def defer_validation
     RULE_2B_2C = /\/([^\/]*)\/\.\.\/|\/([^\/]*)\/\.\.$/
     RULE_2D = /^\.\.?\/?/
     RULE_PREFIXED_PARENT = /^\/\.\.?\/|^(\/\.\.?)+\/?$/
+    HOST_OMITED = false
```
dentarg: This isn't used

lib/addressable/uri.rb
```diff
@@ -2449,10 +2457,8 @@ def validate
         raise InvalidURIError,
           "Absolute URI missing hierarchical segment: '#{self.to_s}'"
       end
-      if self.host == nil
-        if self.port != nil ||
-            self.user != nil ||
-            self.password != nil
+      if host.nil? && @@avoid_host_omitted
```
dentarg: Should this be `@avoid_host_omitted`? `@@avoid_host_omitted` is always `true`

lib/addressable/uri.rb
```diff
@@ -74,6 +74,9 @@ module CharacterClasses
       "ldap" => 389,
       "prospero" => 1525
     }
+    ##
+    #  @@avoid_host_omitted is to avoid host validate when is set to be omitted
+    @@avoid_host_omitted = true
```
dentarg: Why add this?

lib/addressable/uri.rb
JuarezLustosa: Tks fixed

lib/addressable/uri.rb
JuarezLustosa: Because, when the method "omit" is called, the host is nullified, and then call the method  "validate" the error `InvalidURIError, "Hostname not supplied:`. So I don't link this approach to use @@ in this case, but I don't see any solutions to solve this problem without a huge refactory. Do you have any idea?

Slack88 (approved):

spec/addressable/uri_spec.rb
```diff
@@ -1033,6 +1033,19 @@ def to_s
   end
 end
 
+describe Addressable::URI, "when created has a host but " +
```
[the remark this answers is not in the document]
Slack88: Screenshot_2021-09-26-14-30-27-044_net metaquotes metatrader5

CzystyEtyl (approved):

spec/addressable/uri_spec.rb
```diff
@@ -1033,6 +1033,19 @@ def to_s
   end
 end
```
rexingQ, line 1034: https://github.com/sporkmonger/addressable/pull/369#discussion_r353168325

rexingQ (approved):

spec/addressable/uri_spec.rb
```diff
@@ -1033,6 +1033,19 @@ def to_s
...
 
+describe Addressable::URI, "when created has a host but " +
+    "is setup to be omitted" do
+
+  before do
+    @uri = Addressable::URI.parse( "https://user:password@dev.domain.com")
+                           .omit(:scheme, :host, :port, :user, :password)
+  end
+
+  it 'avoid to be raised' do
```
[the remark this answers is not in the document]
rexingQ: - [ ] 

> Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

rexingQ (approved):

rexingQ (approved):

Issue #276, opened 2017-11-14
Addressable::URI.encode_component method is not listed in the doc


ykitamura-mdsol:
The [Addressable::URI.encode_component](https://github.com/sporkmonger/addressable/blob/addressable-2.5.2/lib/addressable/uri.rb#L355) method is not listed in the [Addressable::URI doc](http://www.rubydoc.info/gems/addressable/Addressable/URI).

It seems [this part](https://github.com/sporkmonger/addressable/blob/addressable-2.5.2/lib/addressable/uri.rb#L396-L398) is preventing the method to be listed:
```ruby
    class << self
      alias_method :encode_component, :encode_component
    end
```

I think it should be like this:
```diff
    class << self
-     alias_method :encode_component, :encode_component
+     alias_method :escape_component, :encode_component
    end
```

After making the change and running Yard, the method is listed in the doc:
<img width="500" alt="screen shot 2017-11-14 at 12 42 49 pm" src="https://user-images.githubusercontent.com/16286071/32766101-c17c43ee-c950-11e7-9dc2-9d703b943efd.png">

PR #371, opened 2019-12-23
fixing missing encode component documentation entry

charleystran:
Fixing the documentation to resolve https://github.com/sporkmonger/addressable/issues/276 and add the missing docs entry for Addressable::URI.encode_component

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -421,7 +421,7 @@ module Addressable
     end
 
     class << self
-      alias_method :encode_component, :encode_component
+      alias_method :escape_component, :encode_component
     end
 
     ##


charleystran:
@sporkmonger any clue as to why the 2 jruby and truffelruby builds fail on this PR?  The change is very small and I wouldnt expect it to break the builds. Any thoughts?

dentarg:
@charleystran I guess they are unrelated. [Last green build with truffleruby](https://travis-ci.org/sporkmonger/addressable/jobs/583018680) used 19.2.0 and [the build here](https://travis-ci.org/sporkmonger/addressable/jobs/629520070) uses 19.3.0. Re: JRuby, `bundle install` fails. I suspect this is due to Bundler 2.1.2 is used, which is pretty new, so I rate the chance for a bug somewhere likely.

Feel free to report the errors to the projects, or look into if the CI setup here can be changed in some way to make it work. :)

PR #372, opened 2019-12-26
add ruby 2.7 to ci process

charleystran:


dentarg:
Thanks

(I'll just note that I wrote about the build failures in https://github.com/sporkmonger/addressable/pull/371#issuecomment-569123641)

charleystran:
Thanks!

PR #384, opened 2020-04-16
Test the latest TruffleRuby release

eregon:
cc @dentarg

PR #385, opened 2020-04-16
Test with truffleruby-head

eregon:
Since simplecov needs a fix which is not in TruffleRuby 20.0.0,
see https://github.com/sporkmonger/addressable/pull/384

cc @dentarg


dentarg:
Great :)

PR #387, opened 2020-04-24
fix: avoid some allocations of strings

isqad:
Hello!
In this commit I suggest some optimizations for avoid unnecessary  strings allocations

Before (our loop of mail delivery):
```
21747764  /localgems/addressable/lib/addressable/uri.rb:557
```

After:

no any conversions from nil to empty string

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -73,7 +73,7 @@ module Addressable
       "wais" => 210,
       "ldap" => 389,
       "prospero" => 1525
-    }
+    }.freeze
 
     ##
     # Returns a URI object based on the parsed string.
@@ -463,7 +463,11 @@ module Addressable
       uri = uri.dup
       # Seriously, only use UTF-8. I'm really not kidding!
       uri.force_encoding("utf-8")
-      leave_encoded = leave_encoded.dup.force_encoding("utf-8")
+
+      unless leave_encoded.empty?
+        leave_encoded = leave_encoded.dup.force_encoding("utf-8")
+      end
+
       result = uri.gsub(/%[0-9a-f]{2}/iu) do |sequence|
         c = sequence[1..3].to_i(16).chr
         c.force_encoding("utf-8")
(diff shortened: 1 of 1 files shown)


isqad:
:green_apple: @dentarg

lib/addressable/uri.rb
```diff
@@ -1132,7 +1141,9 @@ def normalized_host
         end
       end
       # All normalized values should be UTF-8
-      @normalized_host.force_encoding(Encoding::UTF_8) if @normalized_host
+      unless @normalized_host.empty?
+        @normalized_host.force_encoding(Encoding::UTF_8)
+      end
```
dentarg, line 1146: Is it guaranteed that `@normalized_host` will always respond to `#empty?`. The previous code guarded against `nil` (Maybe it wasn't necessary or the test coverage we have isn't sufficient, I don't know)

isqad: Ah...ok, but `@normalized_host` can be empty string, so it is proper way?

```ruby
if @normalized_host && !@normalized_host.empty?
  @normalized_host.force_encoding(Encoding::UTF_8)
end
```

dentarg: Yeah

isqad:
:up:

isqad:
:thinking:

sporkmonger:
Merged #383 before this one as requested.

isqad:
Before:
https://travis-ci.org/github/sporkmonger/addressable/jobs/687158167

image
After:
https://travis-ci.org/github/sporkmonger/addressable/jobs/687287165
image

lib/addressable/uri.rb
ashmaroli: What if `@normalized_host` is an empty string but not `UTF-8` encoded..?

isqad: I think that it will not be have matter because url without host is wrong url

:thinking: 

it is may be have sense to add condition on emptiness at this line: https://github.com/sporkmonger/addressable/blob/master/lib/addressable/uri.rb#L1246

ashmaroli:
@dentarg Any updates on this?

ashmaroli:
pinging @dentarg @sporkmonger

PR #389, opened 2020-05-05
Reduce allocations in template

graemeboyd:
I have two commits here for speedups related to matching URIs to templates. I found these by profiling a Middleman static website build. They produce a roughly 2x increase in speed for the use case of matching 10k URIs to the same template.

The first commit avoids a duplication where #match is called with an Addressable::URI object rather than a string. 

The second commit caches the regexp produced by #parse_template_function. When match is called over and over against the same source, generating a new regexp each time is a significant part of the work done.

I did consider a wider refactor e.g. to allow to_regexp to take the optional processor as a parameter or allow the processor to be supplied via an accessor but those would require more extensive changes.

Below is a profile of a match of 10k URIs to one template. The current version (2.7.0) is on the left, this branch on the right.
image

### DIFF
--- a/lib/addressable/template.rb
+++ b/lib/addressable/template.rb
@@ -982,6 +982,10 @@ module Addressable
     # @return [Regexp]
     #   A regular expression which may be used to parse a template pattern.
     def parse_template_pattern(pattern, processor=nil)
+      if processor.nil? && defined?(@expansions) && defined?(@template_pattern_regexp)
+        return [@expansions, @template_pattern_regexp]
+      end
+
       # Escape the pattern. The two gsubs restore the escaped curly braces
       # back to their original form. Basically, escape everything that isn't
       # within an expansion.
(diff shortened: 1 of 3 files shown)


lib/addressable/template.rb
```diff
@@ -1038,7 +1038,18 @@ def parse_template_pattern(pattern, processor=nil)
 
       # Ensure that the regular expression matches the whole URI.
       regexp_string = "^#{regexp_string}$"
-      return expansions, Regexp.new(regexp_string)
+
```
therabidbanana, line 1041: It's been a long time since I looked at this code, but I think the general idea is great! Could probably optimize further while we're at it though.

https://github.com/sporkmonger/addressable/pull/389/files#diff-76daf569c1d16697395cf4be56da5592R984-R987

I feel like we could probably get even more performance wins if we put an early return up towards the top of this function instead - as long as pattern (which is actually immutable, if I remember correctly) and processor (usually nil) passed in matches, we could return a cached version of the expansions and Regexp. Putting the cache at the very bottom of the function reduces Regexp allocations but still lets a lot of cacheable work (which is done by a bunch of String allocations) be done first to compare to the previously cached value. 

One way to approach it might be to rename this old function `parse_fresh_template_pattern` and then make a new `parse_template_pattern` that calls to it only if value isn't already cached for pattern / processor combo. Given processors could be non-deterministic, I might even say only cache if processor is nil?

sporkmonger:
Just merged #383 but I think that only covers URI parsing. Might want to expand that to also profile template methods so that we can more easily compare before/after.

sporkmonger:
This was essentially merged via #398.

Issue #390, opened 2020-05-07
Addressable::URI::InvalidURIError with valid URLs


crondaemon:
Hi
I'm experiencing issues with URLs containing `?//`. Simple reproducer:

```ruby
irb(main):004:0> a = Addressable::URI.parse("http://a.domain.it/?p1=v1&p2=v2")
=> #<Addressable::URI:0x2aeb31315708 URI:http://a.domain.it/?p1=v1&p2=v2>
irb(main):005:0> a.join("/?p1=v1&p2=v2")
=> #<Addressable::URI:0x2aeb313120d0 URI:http://a.domain.it/?p1=v1&p2=v2>
irb(main):006:0> 
```
works as expected. Instead:

```ruby
irb(main):002:0> a = Addressable::URI.parse("http://a.domain.it//?p1=v1&p2=v2")
=> #<Addressable::URI:0x2aeb3135450c URI:http://a.domain.it//?p1=v1&p2=v2>
irb(main):003:0> a.join("//?p1=v1&p2=v2")
Traceback (most recent call last):
        1: from (irb):3
Addressable::URI::InvalidURIError (Absolute URI missing
hierarchical segment: 'http://?p1=v1&p2=v2')
```

If I'm not mistaken the latter is a valid URL.

dentarg:
> If I'm not mistaken the latter is a valid URL.

Exactly what string are you saying is a valid URL?

---

I started looking at the humongous that's `#join`

https://github.com/sporkmonger/addressable/blob/5c04c79cdc8da2de5f15782641a7f102e2627a4b/lib/addressable/uri.rb#L1876-L1975

I could spot this difference that's probably of importance

```ruby
irb(main):024:0> Addressable::URI.parse("//?p1=v1&p2=v2").path
=> ""
irb(main):025:0> Addressable::URI.parse("/?p1=v1&p2=v2").path
=> "/"
```

Feel free to dig in yourself :)

crondaemon:
I'm saying this `http://a.domain.it//?p1=v1&p2=v2` is a valid URL.

crondaemon:
According to this https://www.freeformatter.com/url-parser-query-string-splitter.html the URL should be formally corrent, hence the join should work with both `http://a.domain.it//?p1=v1&p2=v2` and `http://a.domain.it/?p1=v1&p2=v2`.

dentarg:
This works:

```ruby
[...]

PR #391, opened 2020-05-07
Don't use the scheme unless we have a valid host.

crondaemon:
Fix: #390.


sporkmonger:
See #390 for explanation.

Issue #395, opened 2020-07-05
Normalize introduces weird character substitutions.


midnight-wonderer:
I spent a day figuring out what going on with my program and eventually found something weird from `Addressable`.
Please have a look.

By running the following code
~~~ruby
::Addressable::URI.parse('http://example.com/%E0%B8%84%E0%B8%B3%E0%B8%A0%E0%B8%B2%E0%B8%A9%E0%B8%B2%E0%B9%84%E0%B8%97%E0%B8%A2').normalize.to_s
~~~
I expect
~~~ruby
'http://example.com/%E0%B8%84%E0%B8%B3%E0%B8%A0%E0%B8%B2%E0%B8%A9%E0%B8%B2%E0%B9%84%E0%B8%97%E0%B8%A2'
~~~
as an output.

However, the actual output is:
~~~ruby
'http://example.com/%E0%B8%84%E0%B9%8D%E0%B8%B2%E0%B8%A0%E0%B8%B2%E0%B8%A9%E0%B8%B2%E0%B9%84%E0%B8%97%E0%B8%A2'
~~~
Please note the difference in length.

Upon a closer look `%E0%B8%B3` seems to be substituted with `%E0%B9%8D%E0%B8%B2` with no apparent reason.  
Since they are entirely different characters, the substitution does not work.

This is highly unexpected, and I'm not sure if it is the only case. (I only stumble upon one.)

P.S. version 2.7.0

midnight-wonderer:
Digging further the behavior seems to be specified in https://unicode-org.github.io/cldr-staging/charts/37/supplemental/character_fallback_substitutions.html
more specifically, in the NFKC normalization flavor of Unicode standard.

For some context, substitute Thai character ำ with ํ and า is like substitute English character m with r and n, which mostly does not make sense, except they are somewhat look alike.  
Nobody uses ํ anymore in Thai; in fact, it takes me some time to locate the character on my keyboard.  
The only reason for such substitution is for security reasons.
Think, someone can create a seemingly legit domain name imposing as some other legit identities.

Why does Addressable use KC flavor for both domain part and path?  
By looking specifically at this case, KC (NFKC) flavor should be used only for the domain while NFC should be used everywhere else.
From what I read KC is mostly for security reasons, NFC for compatibility. [See FAQ](https://unicode.org/faq/normalization.html)

sporkmonger:
[...]

PR #396, opened 2020-07-05
Address incorrect Unicode normalizations

midnight-wonderer:
The issue is as stated in https://github.com/sporkmonger/addressable/issues/395
This is my proposed changes.

I did not test locally and planing on offloading the work to Travis. I will fix the issues eventually if the tests fail.
In the meantime, I am open to discussions and requests. Feel free to join even if you are not the maintainers of the gem.

### DIFF
--- a/lib/addressable/idna/pure.rb
+++ b/lib/addressable/idna/pure.rb
@@ -117,8 +117,31 @@ module Addressable
     def self.unicode_normalize_kc(input)
       input = input.to_s unless input.is_a?(String)
       unpacked = input.unpack("U*")
-      unpacked =
-        unicode_compose(unicode_sort_canonical(unicode_decompose(unpacked)))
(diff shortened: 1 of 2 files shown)


midnight-wonderer:
Could somebody help to tell me why the build failed on old Ruby versions?  
Do old versions always use native IDN implementations or something?

If it is the case, it will be hard fixing this bug. (I checked the IDN gem, it only provides binding for the `nfkc_normalize` function)

There are lots of real-world pages affected by this bug.  
For instance sites like Wiktionary `https://th.wiktionary.org/wiki/%E0%B8%84%E0%B8%B3`.

sporkmonger:
I'm going to close this and #395 (explanation there), but if this PR is changed to allow path and query normalization to take an optional parameter for which normalization form to use, I'd be fine reopening it. The hostname must always be NFKC and shouldn't be parameterized.

PR #409, opened 2021-02-12
Fix encoding ipv6 literals

### LINKED ISSUE #402: Encoding an IPv6 address raises an exception
Calling `URI.encode` with an IPv6 literal host raises an exception:

```
[1] pry(main)> Addressable::URI.encode("http://[fe80::200:f8ff:fe21:67cf]/")
Addressable::URI::InvalidURIError: Invalid port number: "67cf]"
from /usr/local/lib/ruby/gems/2.6.0/gems/addressable-2.7.0/lib/addressable/uri.rb:1409:in `port='
```
I believe the output here should be `http://[fe80::200:f8ff:fe21:67cf]/`. RFC3986 defines square brackets as [reserved within the authority component](https://tools.ietf.org/html/rfc3986#appendix-D.1), and as such they should not be percent-encoded.

The root cause appears to be https://github.com/sporkmonger/addressable/blob/f839fb28344567c792e148d7a74c529748455d11/lib/addressable/uri.rb#L611-L612

Adding `"\\[\\]"` to `CharacterClasses::AUTHORITY` appears to resolve the issue. The example above returns the expected result.

Square brackets are already included in `CharacterClasses::HOST`. Given that `HOST` is a component of `AUTHORITY` I would naively expect all `HOST` reservations to also be present in `AUTHORITY`, but there may be some nuance I'm missing here.

Very happy to open a PR with that fix, or investigate alternative solutions if needed.

davidtaylorhq:
Previously, this would error with: Addressable::URI::InvalidURIError: Invalid port number

Fixes #402


sporkmonger (approved):
Approved, but let's land #410 first, and make sure it's merged in before proceeding.

sporkmonger:
In the meantime, please update the changelog.

davidtaylorhq:
Sorry for the delay here! Changelog updated, and minor version number bumped (per comment [in the issue](https://github.com/sporkmonger/addressable/issues/402#issuecomment-777878040)). Please let me know if you'd prefer the version bump in a separate commit.

PR #418, opened 2021-07-05
CI: Make tests run in Windows using JRuby 9.2

dentarg:
Workaround for https://github.com/rubygems/rubygems/issues/3256 / https://github.com/jruby/jruby/issues/6745 (sounds like a fix will be shipped in the next JRuby release after 9.2.19.0)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -114,8 +114,7 @@ jobs:
           - { os: windows-2019, ruby: 2.7 }
           - { os: windows-2019, ruby: '3.0' }
           - { os: windows-2019, ruby: jruby-9.1 }
-          # fails with "bundler: command not found: rake"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -114,8 +114,7 @@ jobs:
           - { os: windows-2019, ruby: 2.7 }
           - { os: windows-2019, ruby: '3.0' }
           - { os: windows-2019, ruby: jruby-9.1 }
-          # fails with "bundler: command not found: rake"
-          - { os: windows-2019, ruby: jruby-9.2, allow-failure: true }
```
yurrok, line 118: [](url)

yurrok, line 118: ##- { os: windows-2019, ruby: jruby-9.2, allow-failure: true }

sporkmonger (approved):

PR #431, opened 2021-09-02
Use the whole string instead of a single line for template match

ojab:
Multiline strings shouldn't pass the validation if a single line matches.


ojab:
Marking as ready for review because `First-time contributors need a maintainer to approve running workflows.`

therabidbanana (approved):
Seems like a good change - supporting multi line strings was definitely not intentional 👍

PR #441, opened 2021-11-01
Fixed works with idn-ruby 0.1.3

ukolovda:
(`IDN::Stringprep.nfkc_normalize(str)` returns nil for invalid UTF-8 in latest version 0.1.3).

### DIFF
--- a/lib/addressable/idna/native.rb
+++ b/lib/addressable/idna/native.rb
@@ -31,7 +31,8 @@ module Addressable
      end
 
     def self.unicode_normalize_kc(value)
-      IDN::Stringprep.nfkc_normalize(value.to_s)
+      str = value.to_s
+      IDN::Stringprep.nfkc_normalize(str) || str
     end
 
     def self.to_ascii(value)


dentarg:
From https://github.com/deepfryed/idn-ruby/issues/9:

>In stable Ruby 3.0 it give the error:
>`ArgumentError: NULL pointer given`

Where did we handle that?

ukolovda:
> Where did we handle that?

Sorry, I don't dig into code deep, I don't know. Tests was success with `idn-ruby 0.1.2` but fails with 0.1.3.

dentarg:
> Where did we handle that?

https://github.com/sporkmonger/addressable/blob/f3fb80c0adb60ce7012c36e5d3714b037a280b06/lib/addressable/uri.rb#L581-L589

https://github.com/sporkmonger/addressable/blob/f3fb80c0adb60ce7012c36e5d3714b037a280b06/lib/addressable/uri.rb#L688-L698

dentarg:
No need for this once idn-ruby 0.1.4 is released: deepfryed/idn-ruby@3b4374f

Thanks for looking into this though, @ukolovda!

PR #460, opened 2022-07-24
Loosen public suffix restriction to allow version 5.0

petergoldstein:
public_suffix 5.0 was just released.  It includes updated definitions, and is restricted to Ruby 2.6+.

Loosening this gemspec restriction allows addressable to use the latest version of public_suffix when loaded with a supported Ruby (2.6+).

### DIFF
--- a/addressable.gemspec
+++ b/addressable.gemspec
@@ -24,14 +24,14 @@ Gem::Specification.new do |s|
     s.specification_version = 4
 
     if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
-      s.add_runtime_dependency(%q<public_suffix>.freeze, [">= 2.0.2", "< 5.0"])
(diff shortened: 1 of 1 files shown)


petergoldstein:
CI failures appears to be unrelated, because the same Windows/JRuby 9.3 failures are happening in this PR - https://github.com/sporkmonger/addressable/pull/459 - as well.

dentarg:
Change looks fine to me, no breaking API changes in `public_suffix` what I can see:

* https://github.com/weppos/publicsuffix-ruby/blob/v5.0.0/CHANGELOG.md#500
* https://github.com/weppos/publicsuffix-ruby/compare/v4.0.7...v5.0.0

dentarg:
Totally forgot the fact that there's a rake task to generate the gemspec 🤦

I'll update it

PR #463, opened 2022-07-25
Set required Ruby version to >= 2.2 in gemspec

dentarg:
Addressable 2.8.0 (892861d (part of #416)) dropped support for Ruby 2.0, 2.1.


dentarg:
Totally forgot the fact that there's a rake task to generate the gemspec 🤦

https://github.com/sporkmonger/addressable/pull/464

PR #467, opened 2022-07-25
Magic encoding comment on the first line

### LINKED ISSUE #423: Encoding magic comments are ignored
Comments like https://github.com/sporkmonger/addressable/blob/a079a52ba5a5cb43cadd5edaa684ee6874dafaa8/spec/addressable/idna_spec.rb#L3 (and in other files) might be intended as magic comments for Ruby.
But they have no effect since they are after a blank line.

You can verify by changing it to `# coding: binary`, see it changes nothing, but if you move it as first line then specs break as expected.

It probably still works in practice since UTF-8 is the default for recent Rubies.

Unless maybe such comments are purposefully not meant for Ruby and only editors or so?

Noticed while investigating https://github.com/oracle/truffleruby/issues/2408

dentarg:
...so it works. #309 got this wrong.

See https://ruby-doc.org/core-3.0.0/Encoding.html#class-Encoding-label-Script+encoding

Close #423

### DIFF
--- a/lib/addressable/idna.rb
+++ b/lib/addressable/idna.rb
@@ -1,6 +1,6 @@
+# encoding: UTF-8
 # frozen_string_literal: true
 
-# encoding:utf-8
 #--
 # Copyright (C) Bob Aman
 #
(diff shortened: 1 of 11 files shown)


dentarg:
Yes I was thinking the same thing @houndci-bot 🙈

PR #469, opened 2022-07-25
Update CI matrix

dentarg:


dentarg:
Oh wow, this PR is waiting for status from the jobs I've removed... and the main branch is protected meaning CI needs to be green before I can merge (I'm not admin). It all makes sense that you can not just make a PR removing all the jobs and then do whatever you want... but pretty confusing.

@sporkmonger can you merge?

dentarg:
Everything is green here: https://github.com/sporkmonger/addressable/actions?query=branch%3Aupdate-ci-matrix

PR #481, opened 2022-10-29
Ensure reset of deferred validation

dpep:
I was digging through this code and noticed a few things that could be tightened up (it's generally an awesome library though, so thanks!).  Forgive me if this is unwelcomed feedback.

There is an edge case where the block in defer_validation could raises an exception and validation_deferred isn't reset properly.  Perhaps contrived, but I thought it worth addressing.  The rest are more cosmetic suggestions to make this code simpler / more Rubyish

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -2408,7 +2408,8 @@ module Addressable
       yield
       @validation_deferred = false
       validate
-      return nil
+    ensure
+      @validation_deferred = false
     end
 
   protected
--- a/spec/addressable/uri_spec.rb
+++ b/spec/addressable/uri_spec.rb
@@ -6755,7 +6755,7 @@ describe Addressable::URI, "when deferring validation" do
     end
   end
 
-  it "always resets deferal afterward" do
+  it "always resets deferral afterward" do
     expect { uri.defer_validation { raise "boom" } }.to raise_error("boom")
     expect(deferred).to be false
   end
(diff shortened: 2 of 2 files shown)


lib/addressable/uri.rb
```diff
@@ -851,7 +849,8 @@ def initialize(options={})
         self.query_values = options[:query_values] if options[:query_values]
         self.fragment = options[:fragment] if options[:fragment]
       end
-      self.to_s
+
+      to_s # enforce path validation
```
dpep, line 853: this line took me a bit to figure out, so I figured it was comment worthy.    based on the specs, it looks like `initialize` leverages the validation of `self.path` done in `.to_s`.  but moving it into `.validate` causes other test failures...

dentarg:
>The rest are more cosmetic suggestions to make this code simpler / more Rubyish

Can you put those in a separate PR?

dpep:
@dentarg : spun out into https://github.com/sporkmonger/addressable/pull/482 per your request

spec/addressable/uri_spec.rb
```diff
@@ -6743,3 +6743,25 @@ def to_str
...
+
+  let(:uri) { Addressable::URI.parse("http://example.com") }
+
+  it "defers validation within the block" do
+    uri.defer_validation do
+      expect(deferred).to be true
+    end
+  end
+
+  it "always resets deferal afterward" do
```
dentarg, line 6758:
```suggestion
  it "always resets deferral afterward" do
```

PR #482, opened 2022-10-30
more rubyish

dpep:
while working on https://github.com/sporkmonger/addressable/pull/481 I noticed a few things that could be tightened up and be made more Rushish.  there should be no functionality changes.  Forgive me if this is unwelcomed feedback

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -117,7 +117,7 @@ module Addressable
         uri = uri.to_str
       rescue TypeError, NoMethodError
         raise TypeError, "Can't convert #{uri.class} into String."
-      end if not uri.is_a? String
+      end unless uri.is_a?(String)
 
       # This Regexp supplied as an example in RFC 3986, and it works great.
       scan = uri.scan(URIREGEX)
@@ -138,15 +138,15 @@ module Addressable
           user = userinfo.strip[/^([^:]*):?/, 1]
           password = userinfo.strip[/:(.*)$/, 1]
         end
+
         host = authority.sub(
           /^([^\[\]]*)@/, EMPTY_STR
         ).sub(
           /:([^:@\[\]]*?)$/, EMPTY_STR
         )
+
         port = authority[/:([^:@\[\]]*?)$/, 1]
-      end
-      if port == EMPTY_STR
-        port = nil
+        port = nil if port == EMPTY_STR
       end
 
       return new(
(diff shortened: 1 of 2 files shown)


lib/addressable/uri.rb
```diff
@@ -851,7 +849,8 @@ def initialize(options={})
         self.query_values = options[:query_values] if options[:query_values]
         self.fragment = options[:fragment] if options[:fragment]
       end
-      self.to_s
+
+      to_s # run path validation
```
dpep, line 853: this line took me a bit to figure out, so I figured it was comment worthy. based on the specs, it looks like `initialize` leverages the validation of `self.path` done in `.to_s`. but moving it into `.validate` causes other test failures...

dentarg:
Could give this PR a better title?

dpep:
for sure.  sorry, that got copied over from the other branch.  better?  (otherwise, open to suggestions)

lib/addressable/uri.rb
```diff
@@ -144,9 +144,7 @@ def self.parse(uri)
           /:([^:@\[\]]*?)$/, EMPTY_STR
         )
         port = authority[/:([^:@\[\]]*?)$/, 1]
-      end
-      if port == EMPTY_STR
-        port = nil
+        port = nil if port == EMPTY_STR
```
dentarg, line 147: Looks like this assignment was moved inside `if authority != nil` where it wasn't before. 
Not sure it matters (all tests still pass, but we might lack coverage).

Spotted this when looking at the files instead of the diff.

### This PR

https://github.com/sporkmonger/addressable/blob/7379545925e2a994dadb6a21a5f9a12051a115fd/lib/addressable/uri.rb#L134-L148

### `main`

https://github.com/sporkmonger/addressable/blob/8657465b438d93c707aff4346fdcf74c742b4a53/lib/addressable/uri.rb#L134-L150

dpep: correct.  port is set to `nil` on line 133 before the block, so this check only needs to happen if the authority block changes the port value

dpep: added a test to ensure proper functionality

dentarg:
@dpep sorry letting this drag on... https://github.com/sporkmonger/addressable/pull/486 changed things related to this PR. Does the changes here still make sense after #486? Feel free to rebase this PR and I'll have a look again

dpep:
rebased.  I think these changes still improve code quality / readability, but feel free to close

PR #485, opened 2022-11-27
Add the `Addressable::URI()` method.

postmodern:
Add the `Addressable::URI()` method similar to the `URI()` method.

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -2557,4 +2557,23 @@ module Addressable
       end
     end
   end
+
+  ##
+  # Coerces a URI object into an {Addressable::URI} object.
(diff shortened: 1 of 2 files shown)


dentarg:
What is the motivation behind this? Besides Ruby having `URI()`.

Not sure this idea is a given, as there's also `#heuristic_parse`. This makes it harder to understand which method is really used.

postmodern:
@dentarg the ability to both parse `Strings` and pass `Addressable::URI` objects through. I noticed that `Addressable::URI.parse` does accept `Addressable::URI` objects, but it copies the obect. `Addressable::URI()` does not copy existing `Addressable::URI` objects.

PR #486, opened 2022-12-15
Improve cache hits and JIT friendliness

tenderlove:
Hi,

This PR adds a few commits to improve inline cache hits for instance variables as well as improve JIT friendliness.  Removing and adding instance variables is very bad for JIT compilers, so this PR removes the "remove" calls.  When we want to remove any caches or precomputed values, we simply set the instance variable to `nil`, or to the sentinel object.  We don't have to check whether or not the IV is defined before setting the value, we can just set it.

This PR also removes some calls to `defined?` that didn't do anything.  `defined?` will always return true for local variables, so there was no need to check it.

Finally, since we can stop checking `defined?` on IVs, I was able to convert some of the reader methods to just use `attr_reader`.

Thanks!

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -1285,10 +1285,10 @@ module Addressable
       end
 
       # Password assigned first to ensure validity in case of nil
-      self.password = defined?(new_password) ? new_password : nil
-      self.user = defined?(new_user) ? new_user : nil
-      self.host = defined?(new_host) ? new_host : nil
-      self.port = defined?(new_port) ? new_port : nil
+      self.password = new_password
+      self.user = new_user
+      self.host = new_host
+      self.port = new_port
 
       # Reset dependent values
       @userinfo = nil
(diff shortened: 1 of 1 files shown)


tenderlove:
I'm happy to fix the style issues, but I believe they were in the pre-existing code.  Please let me know, and I'll fix them. 😃

dentarg:
That's fine, it was obvious to me that you aimed for the minimal amount of changes. That made it easy to read the diff. (It is a bit unfortunate that there's a bot with rules that aren't consistent with the existing code; should be addressed in a separate PR IMHO (not saying it is anything you need to do))

I'm thinking about the change from guards/early returns, many lines changed just because they needed indenting. Could that be avoided or is this how it need to be in order to be performant?

Just trying to keep future `git blame`:ing in mind when merging changes.

tenderlove:
In retrospect, I probably should have done the `return unless` pattern to get rid if the whitespace.  I will do it and push again.  Thanks for your patience!

tenderlove:
Ok, the diff should be much smaller now.  Thank you!

dentarg:
I know about `?w=1` (it is in the UI too now) and I did use it :-) 

Thanks for these changes, they look good to me!

dentarg (approved):
I wont squash merge this as the individual commits look to be valuable. Thanks again.

tenderlove:
Thank you!

PR #489, opened 2022-12-30
CI: don't allow truffleruby-21.3 to fail

dentarg:
https://github.com/ruby/setup-ruby/issues/433 was closed with a fix


dentarg:
It was not fixed and will not be fixed https://github.com/ruby/setup-ruby/issues/433#issuecomment-1367991816

Issue #408, opened 2021-02-09
Normalization differences between IDNA::Native and IDNA::Pure


brasic:
Hello and thanks for your work building and maintaining this useful library!

We use Addressable and `IDNA::Pure` at GitHub for a number of URL parsing and generating tasks.  The pure ruby IDN implementation is a bottleneck in some areas (see https://github.com/sporkmonger/addressable/pull/407) so we are currently evaluating a switch to libidn via `IDNA::Native`.  Our test suite found a few interesting differences between the two implementations when it comes to path normalization of percent-encoded `NUL` bytes.  Here's an example:

```ruby
Addressable::URI.parse("http://github.com/foo/bar/.%00./lol").normalized_path # Addressable::IDNA::Pure
# => "/foo/bar/.%00./lol"

Addressable::URI.parse("http://github.com/foo/bar/.%00./lol").normalized_path # Addressable::IDNA::Native
# => "/foo/bar/lol"
```

The behavior change is ultimately due to the following lower-level difference:

```ruby
irb(main):004:0> Addressable::IDNA.unicode_normalize_kc(".\u0000.") # libidn
=> "."
irb(main):006:0> Addressable::IDNA.unicode_normalize_kc(".\u0000.") # pure
=> ".\u0000."
```

Unfortunately in our testing it seems browsers are split on which is the right way to deal with NUL bytes.  [RFC3986](https://tools.ietf.org/html/rfc3986#section-7.3) has a discussion of `%00` but leaves it up to the application (emphasis mine):
>    Percent-encoded octets must be decoded at some point during the dereference process.  Applications must split the URI into its components and subcomponents prior to decoding the octets, as otherwise the decoded octets might be mistaken for delimiters. Security checks of the data within a URI should be applied after decoding the octets.  **Note, however, that the "%00" percent-encoding (NUL) may require special handling and should be rejected if the application is not expecting to receive raw data within a component.**

Are you interested in harmonizing this difference in normalization between the two IDNA backends and which do you think is the appropriate behavior?

sporkmonger:
These should definitely be made consistent. Since we likely can't influence the behavior of `IDNA::Native`, I'm inclined to believe it should be considered the canonical result where behavior is undefined.

brasic:
@sporkmonger gentle ping.  Have you given any more thought to this?  If you're interested in a PR to resolve the discrepancies we found between `IDNA::Pure` and `IDNA::Native` I'm happy to open one, but IMO it would be a major version bump.

[...]

Issue #491, opened 2023-02-07
Improve pure ruby IDNA implementation to match brother behaviors (IDNA2008 and UTS#46)


jarthod:
Forked from #408, this is the separate ticket to deal with the improvement of the default (pure ruby) IDNA implementation.

- **IDNA2003** is the older standard (which was design for unicode 3.2), it's pretty permissive.
- **IDNA2008** is the _newer_ one, which supports all unicode versions up to now (more characters supported, but also makes a lot of string invalid while IDNA2003 was pretty permissive, to reduce confusion/phishing risk and stuff like that).
- **UTS#46** is some kind of standard character mapping used in conjuction with IDNA2008, in order to avoid breaking compatibility with older domains to make the transision smoother.
- There are unfortunately [4 "Deviations" chars](https://unicode.org/reports/tr46/#Table_Deviation_Characters) between IDNA2003 and IDNA2008 which can't be smoothly transitionned, they are simply converted differently in both version. So clients using IDNA2008+UTS#46 have to choose between the old (called "Transitional") and new (called "Non-Transitional") behavior.

AFAICS the expectation is that while all registrar are upgrading to IDNA2008 and only allow valid hostnames (=approximately forever), browsers and web clients in general are encouraged to use IDNA2008+UTS#46 to widen support. So basically unless you're running a registrar, IDNA2008+UTS#46 is the target.

That's why `libidn2` is implementing [IDNA2008 + UTS#46](https://gitlab.com/libidn/libidn2) and default to the **Non-Transitional** (=new) mode. Which is also used by curl for example and probably many other web clients. Firefox and Safari also seems to do IDNA2008+UTS#46 Non-Transitional. Chrome was lagging a bit as it was still using Transitional mode up until very recently, apparently [they juuust changed this to Non-Transitional in Chome 110](https://chromestatus.com/feature/5105856067141632). I can't verify this yet as I only have Chome 109 on Linux ^^

`libidn` (the current "native" option) implements **IDNA2003** standard (the "older" one). IMO we should upgrade to `libidn2`, this will be discussed in  #247.

The "pure" implementation is **IDNA2008iiiisssshhhhh**, but not compliant. As we can see in this example with an emoji modifier:
```ruby
irb(main):004:0> s1 = "https://l♥️h.ws"
=> "https://l♥️h.ws"
irb(main):006:0> Addressable::URI.parse(s1).normalize
[...]

PR #449, opened 2022-01-15
Explicitly freeze concatenated-`String` constants to unbreak on non-main `Ractor`s.

okeeblow:
When building `String`s with the `String#concat` / `String#+` methods, the receiver and argument `String`s are literals, but the resulting value is not, so they are left unfrozen when we rely only on the `frozen_string_literal` directive.

Switch to the [`%q` non-interpolable `String` literal syntax](https://docs.ruby-lang.org/en/master/doc/syntax/literals_rdoc.html#label-25q-3A+Non-Interpolable+String+Literals) for these because it feels less gross to me than e.g. `HOST = (UNRESERVED + SUB_DELIMS + "\\[\\:\\]").freeze`. This syntax [was present in Ruby 2.0](https://docs.ruby-lang.org/en/2.0.0/syntax/literals_rdoc.html#label-Percent+Strings), which is Addressable's minimum requirement :)

## Before
Concatenated `String`s are not frozen when the `frozen_string_literal: true` directive is used.

```
[okeeblow@emi#CHECKING YOU OUT] ruby2.7 -r 'addressable' -e 'Addressable::URI::CharacterClasses::constants::each { p "#{_1} frozen? #{Addressable::URI::CharacterClasses::const_get(_1).frozen?}" }'
"ALPHA frozen? true"
"DIGIT frozen? true"
"GEN_DELIMS frozen? true"
"SUB_DELIMS frozen? true"
"RESERVED frozen? false"
"UNRESERVED frozen? false"
"PCHAR frozen? false"
"SCHEME frozen? false"
"HOST frozen? false"
"AUTHORITY frozen? false"
"PATH frozen? false"
"QUERY frozen? false"
"FRAGMENT frozen? false"
```

…and we encounter a `Ractor::IsolationError` when trying to use `Addressable::URI::parse` in a non-main `Ractor`:

```
[okeeblow@emi#CHECKING YOU OUT] ruby -v
ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-linux]

[okeeblow@emi#CHECKING YOU OUT] ruby -r 'addressable' -e 'Addressable::URI::CharacterClasses::constants::each { value = Addressable::URI::CharacterClasses::const_get(_1); p "#{_1} frozen? #{_1.frozen?}; shareable? #{Ractor::shareable?(value)}"}'
"ALPHA frozen? true; shareable? true"
"DIGIT frozen? true; shareable? true"
"GEN_DELIMS frozen? true; shareable? true"
[...]

PR #438, opened 2021-10-08
Remove redundant colon in Addressable::URI::CharacterClasses::AUTHORITY

ukolovda:
Addressed to #437

PR #492, opened 2023-02-13
Use ruby unicode normalize to avoid libidn C problems and heavy legacy ruby code

jarthod:
This is a fix for #408 and the beginning of the simplification and modernization of the IDNA code. It does:
- Replaces `libidn` unicode normalize implementation by ruby's, to avoid the null terminated string problem and support more recent unicode versions.
- Replaces `Pure` unicode normalize by ruby's, to avoid hundreds of line of slow and legacy ruby code and support more recent unicode versions.

The result in terms of performance is the following (benchmark code commited, but needs to be run on master to measure the legacy "pure" code):
```
#              user     system      total        real
# pure     1.325948   0.000000   1.325948 (  1.326054)
# libidn   0.058067   0.000000   0.058067 (  0.058069)
# ruby     0.325062   0.000000   0.325062 (  0.325115)
```

The native ruby implementation is 5-6 times **slower** than `libidn`, but also 5 times **faster** than the existing `Pure` one.

Considering this is just a small part and the rest of the ruby code is more than one order of magnitude slower (doing a full `normalize` in the `simple.rb` benchmark for example takes 8.6 seconds for the same 100k iterations on my machine), I think this is more than enough, no need to try to shave off microseconds in the `unicode_normalize` code if the rest is a lot of ruby.

And we benefit from the native ruby function which is going to be maintained and improved.
Specs have been added to cover the problematic `\x00` case.

### DIFF
--- a/benchmark/unicode_normalize.rb
+++ b/benchmark/unicode_normalize.rb
@@ -1,0 +1,34 @@
+# /usr/bin/env ruby
+# frozen_string_literal: true.
+
+require "benchmark"
+require "addressable/idna/pure.rb"
+require "idn"
+
+value = "ﬁﾯリ宠퐱卄.com"
+expected = "fiᆵリ宠퐱卄.com"
+N = 100_000
+
+fail "ruby does not match" unless expected == value.unicode_normalize(:nfkc)
+fail "libidn does not match" unless expected == IDN::Stringprep.nfkc_normalize(value)
+fail "addressable does not match" unless expected == Addressable::IDNA.unicode_normalize_kc(value)
+
+Benchmark.bmbm do |x|
+  x.report("pure") { N.times { Addressable::IDNA.unicode_normalize_kc(value) } }
+  x.report("libidn") { N.times { IDN::Stringprep.nfkc_normalize(value) } }
+  x.report("ruby") { N.times { value.unicode_normalize(:nfkc) } }
+end
+
+# February 14th 2023, before replacing the legacy pure normalize code:
+
+# > ruby benchmark/unicode_normalize.rb
+# Rehearsal ------------------------------------------
+# pure     1.335230   0.000315   1.335545 (  1.335657)
+# libidn   0.058568   0.000000   0.058568 (  0.058570)
+# ruby     0.326008   0.000014   0.326022 (  0.326026)
+# --------------------------------- total: 1.720135sec
+
+#              user     system      total        real
+# pure     1.325948   0.000000   1.325948 (  1.326054)
+# libidn   0.058067   0.000000   0.058067 (  0.058069)
+# ruby     0.325062   0.000000   0.325062 (  0.325115)
--- a/lib/addressable/idna/native.rb
+++ b/lib/addressable/idna/native.rb
@@ -29,10 +29,6 @@ module Addressable
        IDN::Punycode.decode(value.to_s)
      end
 
-    def self.unicode_normalize_kc(value)
-      IDN::Stringprep.nfkc_normalize(value.to_s)
-    end
-
     def self.to_ascii(value)
       value.to_s.split('.', -1).map do |segment|
         if segment.size > 0 && segment.size < 64
--- a/lib/addressable/idna/pure.rb
+++ b/lib/addressable/idna/pure.rb
@@ -66,7 +66,7 @@ module Addressable
     # domain name as described in RFC 3490.
     def self.to_ascii(input)
       input = input.to_s unless input.is_a?(String)
-      input = input.dup
+      input = input.dup.force_encoding(Encoding::UTF_8).unicode_normalize(:nfkc)
       if input.respond_to?(:force_encoding)
         input.force_encoding(Encoding::ASCII_8BIT)
       end
@@ -77,7 +77,7 @@ module Addressable
             part.force_encoding(Encoding::ASCII_8BIT)
           end
           if part =~ UTF8_REGEX && part =~ UTF8_REGEX_MULTIBYTE
-            ACE_PREFIX + punycode_encode(unicode_normalize_kc(part))
+            ACE_PREFIX + punycode_encode(part)
           else
             part
           end
@@ -112,15 +112,6 @@ module Addressable
       output
     end
 
-    # Unicode normalization form KC.
-    def self.unicode_normalize_kc(input)
-      input = input.to_s unless input.is_a?(String)
-      unpacked = input.unpack("U*")
-      unpacked =
-        unicode_compose(unicode_sort_canonical(unicode_decompose(unpacked)))
-      return unpacked.pack("U*")
-    end
-
     ##
     # Unicode aware downcase method.
     #
@@ -136,164 +127,6 @@ module Addressable
     end
     private_class_method :unicode_downcase
 
-    def self.unicode_compose(unpacked)
-      unpacked_result = []
-      length = unpacked.length
-
-      return unpacked if length == 0
-
-      starter = unpacked[0]
-      starter_cc = lookup_unicode_combining_class(starter)
-      starter_cc = 256 if starter_cc != 0
-      for i in 1...length
-        ch = unpacked[i]
-
-        if (starter_cc == 0 &&
-            (composite = unicode_compose_pair(starter, ch)) != nil)
-          starter = composite
-        else
-          unpacked_result << starter
-          starter = ch
-        end
-      end
-      unpacked_result << starter
-      return unpacked_result
-    end
-    private_class_method :unicode_compose
-
-    def self.unicode_compose_pair(ch_one, ch_two)
-      if ch_one >= HANGUL_LBASE && ch_one < HANGUL_LBASE + HANGUL_LCOUNT &&
-          ch_two >= HANGUL_VBASE && ch_two < HANGUL_VBASE + HANGUL_VCOUNT
-        # Hangul L + V
-        return HANGUL_SBASE + (
-          (ch_one - HANGUL_LBASE) * HANGUL_VCOUNT + (ch_two - HANGUL_VBASE)
-        ) * HANGUL_TCOUNT
-      elsif ch_one >= HANGUL_SBASE &&
-          ch_one < HANGUL_SBASE + HANGUL_SCOUNT &&
-          (ch_one - HANGUL_SBASE) % HANGUL_TCOUNT == 0 &&
-          ch_two >= HANGUL_TBASE && ch_two < HANGUL_TBASE + HANGUL_TCOUNT
-           # Hangul LV + T
-        return ch_one + (ch_two - HANGUL_TBASE)
-      end
-
-      p = []
-
-      ucs4_to_utf8(ch_one, p)
-      ucs4_to_utf8(ch_two, p)
-
-      return lookup_unicode_composition(p)
-    end
-    private_class_method :unicode_compose_pair
-
-    def self.ucs4_to_utf8(char, buffer)
-      if char < 128
-        buffer << char
-      elsif char < 2048
-        buffer << (char >> 6 | 192)
-        buffer << (char & 63 | 128)
-      elsif char < 0x10000
-        buffer << (char >> 12 | 224)
-        buffer << (char >> 6 & 63 | 128)
-        buffer << (char & 63 | 128)
-      elsif char < 0x200000
-        buffer << (char >> 18 | 240)
-        buffer << (char >> 12 & 63 | 128)
-        buffer << (char >> 6 & 63 | 128)
-        buffer << (char & 63 | 128)
-      elsif char < 0x4000000
-        buffer << (char >> 24 | 248)
-        buffer << (char >> 18 & 63 | 128)
-        buffer << (char >> 12 & 63 | 128)
-        buffer << (char >> 6 & 63 | 128)
-        buffer << (char & 63 | 128)
-      elsif char < 0x80000000
-        buffer << (char >> 30 | 252)
-        buffer << (char >> 24 & 63 | 128)
-        buffer << (char >> 18 & 63 | 128)
-        buffer << (char >> 12 & 63 | 128)
-        buffer << (char >> 6 & 63 | 128)
-        buffer << (char & 63 | 128)
-      end
-    end
-    private_class_method :ucs4_to_utf8
-
-    def self.unicode_sort_canonical(unpacked)
-      unpacked = unpacked.dup
-      i = 1
-      length = unpacked.length
-
-      return unpacked if length < 2
-
-      while i < length
-        last = unpacked[i-1]
-        ch = unpacked[i]
-        last_cc = lookup_unicode_combining_class(last)
-        cc = lookup_unicode_combining_class(ch)
-        if cc != 0 && last_cc != 0 && last_cc > cc
-          unpacked[i] = last
-          unpacked[i-1] = ch
-          i -= 1 if i > 1
-        else
-          i += 1
-        end
-      end
-      return unpacked
-    end
-    private_class_method :unicode_sort_canonical
-
-    def self.unicode_decompose(unpacked)
-      unpacked_result = []
-      for cp in unpacked
-        if cp >= HANGUL_SBASE && cp < HANGUL_SBASE + HANGUL_SCOUNT
-          l, v, t = unicode_decompose_hangul(cp)
-          unpacked_result << l
-          unpacked_result << v if v
-          unpacked_result << t if t
-        else
-          dc = lookup_unicode_compatibility(cp)
-          unless dc
-            unpacked_result << cp
-          else
-            unpacked_result.concat(unicode_decompose(dc.unpack("U*")))
-          end
-        end
-      end
-      return unpacked_result
-    end
-    private_class_method :unicode_decompose
-
-    def self.unicode_decompose_hangul(codepoint)
-      sindex = codepoint - HANGUL_SBASE;
-      if sindex < 0 || sindex >= HANGUL_SCOUNT
-        l = codepoint
-        v = t = nil
-        return l, v, t
-      end
-      l = HANGUL_LBASE + sindex / HANGUL_NCOUNT
-      v = HANGUL_VBASE + (sindex % HANGUL_NCOUNT) / HANGUL_TCOUNT
-      t = HANGUL_TBASE + sindex % HANGUL_TCOUNT
-      if t == HANGUL_TBASE
-        t = nil
-      end
-      return l, v, t
-    end
-    private_class_method :unicode_decompose_hangul
-
-    def self.lookup_unicode_combining_class(codepoint)
-      codepoint_data = UNICODE_DATA[codepoint]
-      (codepoint_data ?
-        (codepoint_data[UNICODE_DATA_COMBINING_CLASS] || 0) :
-        0)
-    end
-    private_class_method :lookup_unicode_combining_class
-
-    def self.lookup_unicode_compatibility(codepoint)
-      codepoint_data = UNICODE_DATA[codepoint]
-      (codepoint_data ?
-        codepoint_data[UNICODE_DATA_COMPATIBILITY] : nil)
-    end
-    private_class_method :lookup_unicode_compatibility
-
     def self.lookup_unicode_lowercase(codepoint)
       codepoint_data = UNICODE_DATA[codepoint]
       (codepoint_data ?
@@ -301,21 +134,6 @@ module Addressable
         codepoint)
     end
     private_class_method :lookup_unicode_lowercase
-
-    def self.lookup_unicode_composition(unpacked)
-      return COMPOSITION_TABLE[unpacked]
-    end
-    private_class_method :lookup_unicode_composition
-
-    HANGUL_SBASE =  0xac00
-    HANGUL_LBASE =  0x1100
-    HANGUL_LCOUNT = 19
-    HANGUL_VBASE =  0x1161
-    HANGUL_VCOUNT = 21
-    HANGUL_TBASE =  0x11a7
-    HANGUL_TCOUNT = 28
-    HANGUL_NCOUNT = HANGUL_VCOUNT * HANGUL_TCOUNT # 588
-    HANGUL_SCOUNT = HANGUL_LCOUNT * HANGUL_NCOUNT # 11172
 
     UNICODE_DATA_COMBINING_CLASS = 0
     UNICODE_DATA_EXCLUSION = 1
--- a/lib/addressable/template.rb
+++ b/lib/addressable/template.rb
@@ -892,7 +892,7 @@ module Addressable
     # operator.
     #
     # @param [Hash, Array, String] value
-    #   Normalizes keys and values with IDNA#unicode_normalize_kc
+    #   Normalizes unicode keys and values with String#unicode_normalize (NFC)
     #
     # @return [Hash, Array, String] The normalized values
     def normalize_value(value)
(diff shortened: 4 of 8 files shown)


lib/addressable/idna/pure.rb
```diff
@@ -77,7 +79,7 @@ def self.to_ascii(input)
             part.force_encoding(Encoding::ASCII_8BIT)
           end
           if part =~ UTF8_REGEX && part =~ UTF8_REGEX_MULTIBYTE
-            ACE_PREFIX + punycode_encode(unicode_normalize_kc(part))
+            ACE_PREFIX + punycode_encode(part)
```
jarthod, line 82: And we need to do the normalise while we have an unicode string so before the conversion to `ASCII_8BIT`
Also the normalize should be before the split according to http://www.unicode.org/reports/tr46/#ProcessingStepNormalize.
So in the end it's both simpler and more correct this way :)

lib/addressable/uri.rb
```diff
@@ -53,7 +53,7 @@ module CharacterClasses
       PCHAR = (UNRESERVED + SUB_DELIMS + "\\:\\@").freeze
       SCHEME = (ALPHA + DIGIT + "\\-\\+\\.").freeze
       HOST = (UNRESERVED + SUB_DELIMS + "\\[\\:\\]").freeze
-      AUTHORITY = (PCHAR + "\\[\\:\\]").freeze
+      AUTHORITY = (PCHAR + "\\[\\]").freeze
```
jarthod, line 56: Unrelated minor fix to remove the following warning: `uri.rb:417: warning: character class has duplicated range: /[^a-zA-Z0-9\-\.\_\~\!\$\&\'\(\)\*\+\,\;\=\:\@\[\:\]]` because the column `:` is already in `PCHAR`

spec/addressable/idna_spec.rb
```diff
@@ -38,6 +38,12 @@
     )).to eq("www.xn--8ws00zhy3a.com")
   end
 
+  it "also accepts unicode strings encoded as ascii-8bit" do
+    expect(Addressable::IDNA.to_ascii(
+      "www.詹姆斯.com".b
+    )).to eq("www.xn--8ws00zhy3a.com")
+  end
+
```
jarthod, line 46: The ruby normalize function requires a proper Unicode string which was not the case before, so even though this was not the case anywhere, I though I would add a spec to make sure it still works even if the user (or other code) passes a wrongly encoded string (`ASCII-8BIT` in this example thanks to `.b`)

spec/addressable/idna_spec.rb
```diff
@@ -253,6 +259,7 @@
   it "should normalize 'string' correctly" do
     expect(Addressable::IDNA.unicode_normalize_kc(:'string')).to eq("string")
     expect(Addressable::IDNA.unicode_normalize_kc("string")).to eq("string")
+    expect(Addressable::IDNA.unicode_normalize_kc("str\x00ing")).to eq("str\x00ing")
```
jarthod, line 262: Fails with `Native` before the fix.

spec/addressable/idna_spec.rb
```diff
@@ -38,6 +38,12 @@
     )).to eq("www.xn--8ws00zhy3a.com")
   end
 
+  it "also accepts unicode strings encoded as ascii-8bit" do
+    expect(Addressable::IDNA.to_ascii(
+      "www.詹姆斯.com".b
```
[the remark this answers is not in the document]
jarthod: This does not match any line in this file

lib/addressable/uri.rb
dentarg: Isn't this already done in https://github.com/sporkmonger/addressable/pull/438?

jarthod: Ah indeed, I must have forked from and outdated local copy, thanks for pointing this out I'll rebase :)

jarthod: Actually I was up to date, the column is back and I'm not even sure why :thinking:. I checked the git commit log, I can see this commit which remove the column: dd84176c2ad6c71bbec4b66bfc694d80969149ae and then this commit: 178bcd97fc77b342a3bda6f5b6e17f8ea344bd51 added a `freeze` but somehow the column is back, even in the "previous" version in red...

`git log -p lib/addressable/uri.rb`:
image

I dont... :shrug: Any idea how this could happen? Anyway I think we can keep this change to put it back in the codebase then.

benchmark/unicode_normalize.rb
```diff
@@ -0,0 +1,34 @@
...
+require "addressable/idna/pure.rb"
+require "idn"
+
+value = "ﬁﾯリ宠퐱卄.com"
+expected = "fiᆵリ宠퐱卄.com"
+N = 100_000
+
+fail "ruby does not match" unless expected == value.unicode_normalize(:nfkc)
+fail "libidn does not match" unless expected == IDN::Stringprep.nfkc_normalize(value)
+fail "addressable does not match" unless expected == Addressable::IDNA.unicode_normalize_kc(value)
```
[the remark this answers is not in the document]
jarthod: Just a benchmark file and it's more readable this way IMO, but let me know if you really want to follow

lib/addressable/template.rb
```diff
@@ -902,15 +902,14 @@ def normalize_value(value)
...
         value = value.inject({}) { |acc, (k, v)|
-          acc[Addressable::IDNA.unicode_normalize_kc(k)] =
-            Addressable::IDNA.unicode_normalize_kc(v)
+          acc[normalize_value(k)] = normalize_value(v)
           acc
         }
-      else
-        value = Addressable::IDNA.unicode_normalize_kc(value)
+      elsif value.encoding == Encoding::UTF_8
+        value = value.unicode_normalize(:nfc)
```
jarthod, line 912: According to [the RFC](https://www.rfc-editor.org/rfc/rfc6570.txt), this normalization is actually supposed to be NFC form, not NFKC:
> If a value is provided by a user, such as via a data-entry dialog, then the string SHOULD be normalized as Normalization Form C (NFC: Canonical Decomposition, followed by Canonical Composition) prior to being used in expansions by a template processor.

It's easy to do now with ruby so instead of using the global function I'm switching to direct ruby call. Also I skip the normalize if the input string is not Unicode, this is faster and avoids potential exceptions. Finally I used a recursive call for the array and hash keys/values, this way the same behavior (NFC and only normalize if Unicode) is shared between all types.

lib/addressable/uri.rb
```diff
@@ -481,7 +481,7 @@ def self.unencode(uri, return_type=String, leave_encoded='')
         leave_encoded.include?(c) ? sequence : c
       end
 
-      result.force_encoding("utf-8")
+      result.force_encoding(Encoding::UTF_8)
```
jarthod, line 484: Just for consistency, we use this way in all other places.

lib/addressable/uri.rb
```diff
@@ -687,8 +687,7 @@ def self.normalized_encode(uri, return_type=String)
       components.each do |key, value|
         if value != nil
           begin
-            components[key] =
-              Addressable::IDNA.unicode_normalize_kc(value.to_str)
+            components[key] = value.to_str.unicode_normalize(:nfc)
```
jarthod, line 690: Same thing for the both calls in URI we should use NFC form instead of NFKC (only for IDNA), so I switched to the simple ruby call. `:nfc` is the default form but I let it as an argument to show that it's a deliberate choice.

spec/addressable/idna_spec.rb
```diff
@@ -249,11 +255,6 @@
       "example..host"
     )).to eq("example..host")
   end
-
-  it "should normalize 'string' correctly" do
-    expect(Addressable::IDNA.unicode_normalize_kc(:'string')).to eq("string")
-    expect(Addressable::IDNA.unicode_normalize_kc("string")).to eq("string")
-  end
```
jarthod, line 256: This function is no longer needed and there's no need to test ruby own implementation.

jarthod:
Ok I've fixed some encoding issues with template on ruby 2.7- (passed with some ascii strings it was failing) and most hound warnings. I've left comments to explain the changes. All tests are passing. Let me know if you have any comment.

As another validation step, I'm gonna run this version in my product (https://updown.io) which converts 150k+ URl regularly to see if I spot anything.

lib/addressable/uri.rb
dentarg: I #449 was opened the same day #438 was merged so #449 probably didn't have the latest changes and we did not notice in review 😞

dentarg: Plus GitHub not understanding it was a conflict? Anyway, I asked @sporkmonger to double check repo settings in that area (as I don't have access to that myself)

lib/addressable/template.rb
dentarg: >I skip the normalize if the input string is not Unicode

Is it possible for the input string to be Unicode but not UTF-8? 🤔 (Maybe not a realistic scenario even if it can happen?)

benchmark/unicode_normalize.rb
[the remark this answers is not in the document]
dentarg: I agree with you and I would raise this limit if I could, but let's follow the bot so it doesn't complain in the future.

(Maybe `# rubocop:disable Metrics/LineLength` comments work? (don't forget to enable the cop again after the section where it is needed.)

spec/addressable/idna_spec.rb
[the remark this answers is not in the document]
dentarg: I don't follow, is the complaint from the bot not correct? :)

dentarg (approved):
Looks good to me, though I'm no expert at all in this area! I trust our tests and your testing @jarthod :-) Thank you for this very clear pull request.

spec/addressable/idna_spec.rb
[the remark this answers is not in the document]
jarthod: What I mean is that all the specs in this file (and most in other spec files) are written with this indentation, so they are all "incorrect" for @houndci-bot. So if I change it to folow this recommendation, it'll look unaligned with other tests and waste a lost of _precious_ chars toward the 80 chars limit. So if we try to make it consistent and fix the other specs, we'll have lines "too long" again and will need even more fixes. Overall making the code much harder to read so a bad idea to foolow it IMO.

lib/addressable/template.rb
dentarg: Ah, this is in `Addressable::Template`, maybe @therabidbanana want to chime in?

I'm really not familiar with templates but my feeling is that the change is okay.

benchmark/unicode_normalize.rb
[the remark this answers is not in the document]
dentarg: I don't know. Maybe don't please the bot and I will try my to get access to repo settings, to remove the bot. :)

jarthod:
> Looks good to me, though I'm no expert at all in this area! I trust our tests and your testing @jarthod :-) Thank you for this very clear pull request.

Thanks! @brasic maybe you also have an opinion or feedback on this PR?

> I was thinking about a major version bump, but I guess this very much a bug fix? Unless there are some very strange edge cases we don't know about and people rely on.

I don't think this deserves a major bump, it's more of a bugfix indeed.
#491 and #247 definitely will though :)

spec/addressable/idna_spec.rb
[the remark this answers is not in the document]
dentarg: Thanks for explaining. Best course of action seems to ignore the bot, try to remove it, and if rules are introduced again, also update the code when introducing rules.

dentarg:
> #491 and #247 definitely will though :)

Hehe, yes, was thinking that too :)

dentarg:
> #491 and #247 definitely will though :)

Hehe, yes, was thinking that too :)

spec/addressable/idna_spec.rb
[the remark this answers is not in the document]
jarthod: I believe so too :) for the record I am not opposed to this kind of bot, but I am definitely against applying them "blindly" when it makes the code less readable and/or less consistent.

lib/addressable/uri.rb
jarthod: Weirdest to me is that it's not only Github here, even in plain git (as in my git log screenshot) we can clearly see the "before" (red) line in https://github.com/sporkmonger/addressable/commit/178bcd97fc77b342a3bda6f5b6e17f8ea344bd51 being different than the "after" (green) line in https://github.com/sporkmonger/addressable/commit/dd84176c2ad6c71bbec4b66bfc694d80969149ae. And there's no other comits on this file in between. I have no idea how this can even be possible.

Anyway I suppose it's ok to keep this change here then.

benchmark/unicode_normalize.rb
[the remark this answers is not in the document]
jarthod: Ok let's ignore the bot for now, and if you want I can try to have a look at adding a `.hound.yml` in another PR (to keep it but tune it down). Not needed if you aim at removing it though.

lib/addressable/uri.rb
dentarg: Yeah we'll keep it here 

The merge commit diff is correct: https://github.com/sporkmonger/addressable/commit/7ecf75146c1a53b089150ccc33c909214d728323

jarthod: Interesting, I don't see this commit in the git log for `lib/addressable/uri.rb`: https://github.com/sporkmonger/addressable/commits/main/lib/addressable/uri.rb

And when I check it with:
```bash
> git show 7ecf75146c1a53b089150ccc33c909214d728323
commit 7ecf75146c1a53b089150ccc33c909214d728323
Merge: 068f673 41f12dd
Author: Bob Aman <sporkmonger@users.noreply.github.com>
Date:   Fri Jul 29 20:35:38 2022 -0700

    Merge pull request #449 from okeeblow/freeze_concatenated_strings
    
    Explicitly freeze concatenated-`String` constants to unbreak on non-main `Ractor`s.

>
```
It's "empty", there's no file changed. There's probably many things I still don't get about `git` :smiling_face_with_tear:

dentarg: Yeah, I just [googled this](https://stackoverflow.com/questions/43138569/how-to-get-git-log-p-to-show-changes-in-merge-commits) and I learned something too here :) with `-m` (or the longer variant `--diff-merges=on`) `git log`/`git show` shows me the diff

```
git log -1 --diff-merges=on 7ecf751
git show --diff-merges=on 7ecf751
```

jarthod: Oh interesting, thanks :)

lib/addressable/template.rb
sporkmonger: I'm inclined to agree that this should be changed. It looks like there may be a similar issue with RFC 5890?

> A "U-label" is an IDNA-valid string of Unicode characters, in Normalization Form C (NFC) and including at least one non-ASCII character, expressed in a standard Unicode Encoding Form (such as UTF-8).

A subtle change like this probably deserves some extra attention in tests.

lib/addressable/idna/pure.rb
```diff
@@ -66,7 +66,7 @@ module IDNA
     # domain name as described in RFC 3490.
     def self.to_ascii(input)
       input = input.to_s unless input.is_a?(String)
-      input = input.dup
+      input = input.dup.force_encoding(Encoding::UTF_8).unicode_normalize(:nfkc)
```
sporkmonger, line 69: Should this still be NFKC?

jarthod: Good question and tricky answer. As far as I understand in [IDNA2008+UTC#46](https://unicode.org/reports/tr46/) there is the purely unicode normalization normalization step which is NFC, but there is also the IDNA mapping table, which is based on NFKC Casefold transformation. So in the end it does look similar to NFKC with some variations.

In IDNA2003 the "normalization" step was called "STRINGPREP" and was equivalent to NFKC (https://www.rfc-editor.org/rfc/rfc3491#ref-STRINGPREP):
> This profile specifies using Unicode normalization form KC, as described in [[STRINGPREP](https://www.rfc-editor.org/rfc/rfc3491#ref-STRINGPREP)].

Which is why [the libidn library only provides NFKC normalization functions](https://www.gnu.org/software/libidn/manual/libidn.html#Unicode-Normalization).

So in the end as we don't have the official mapping for the moment, NFKC is closer to the expected result (and was previously used), so I thought it would be better to keep it like that for the moment. This will probably change in #491 once we have the mapping table and all the conformance tests.

lib/addressable/template.rb
jarthod: > I'm inclined to agree that this should be changed

Do you mean changed back to accepting any encoding and treating it as UTF-8? or keeping the change that I made?

> It looks like there may be a similar issue with RFC 5890/5895?

I believe yes if a user passes in a UTC-16 encoded string for domain name, it'll probably raise. If we want to make this more strict in the IDNA part we could accept any input encoding and re-`encode` it as UTF-8 (instead of `force_encoding`). This is much more likely to break existing people code though if there are invalid or inconvertible chars. Probably something we can discuss more in #491

> A subtle change like this probably deserves some extra attention in tests.

Good point I'll add some more tests around this.

jarthod: Ok I've just added two specs covering the new case: non-unicode strings passed which should **not** be normalized or broken. This direct link should work: https://github.com/sporkmonger/addressable/pull/492/files#diff-ebf4cdeb47084eeefe86a7b9e669d9b89adb907a1c5c19c2928233fabb2c4f86

The result is :green_circle: green on this branch.
With `main` branch the result is :red_circle: :
```
  1) Addressable::Template Partial expand with symbols partial expand with unicode values preserves other encodings when input is not unicode
     Failure/Error: unpacked = input.unpack("U*")
     
     ArgumentError:
       malformed UTF-8 character
     # ./lib/addressable/idna/pure.rb:119:in `unpack'
     # ./lib/addressable/idna/pure.rb:119:in `unicode_normalize_kc'
     # ./lib/addressable/template.rb:948:in `normalize_value'
     # ./lib/addressable/template.rb:809:in `block in transform_capture'
     # ./lib/addressable/template.rb:791:in `each'
     # ./lib/addressable/template.rb:791:in `inject'
     # ./lib/addressable/template.rb:791:in `transform_capture'
     # ./lib/addressable/template.rb:750:in `block in transform_partial_capture'
     # ./lib/addressable/template.rb:747:in `each'
     # ./lib/addressable/template.rb:747:in `inject'
     # ./lib/addressable/template.rb:747:in `transform_partial_capture'
     # ./lib/addressable/template.rb:529:in `block in partial_expand'
     # ./lib/addressable/template.rb:528:in `gsub!'
     # ./lib/addressable/template.rb:528:in `partial_expand'
     # ./spec/addressable/template_spec.rb:1017:in `block (4 levels) in <top (required)>'
```

Probably not a big use-case otherwise people would have complained more I suppose? but I thought it was worth "improving" that.

The only case which could be "degraded" by this new code as @dentarg noted is if people pass in actual unicode string wrongly declared as another encoding. In such case they won't be normalized, whereas in the previous code they would have been. I don't think that's big deal as it was kind of a side-effect initially, and restoring this "behavior" will mean breaking compatibility with some other encodings, and for people who are correctly passing their strings encoding on top of it.

@dentarg @sporkmonger does this makes sense?

sporkmonger: Thanks for being thorough and checking the new test against `main`.

sporkmonger: So I think this is the key part of the template spec related to expansion methods:
> It is only during the process of template expansion that a string of characters in a URI Template is REQUIRED to be processed as a sequence of Unicode code points.
I read that as probably disallowing the behavior specified by this test unfortunately?

jarthod: > Thanks for being thorough and checking the new test against main.

:+1:

> I read that as probably disallowing the behavior described by this test description?

Good point, as I quickly scanned the spec (which is new to me) I noticed they mentionned some flexibility ("it should be understood that templates occur in practice as a sequence of characters in whatever form or encoding is suitable for the context in which they occur") which is what I expected considering the world of URI so I stopped there. And I missed the part you're showing right after that which says that for **expansion**, unicode is required.

In that case it makes things simpler, so I changed the code to accept any string, force encode it as UTF-8 and normalize it (so similar to `main` behavior). I've updated the specs to match this: https://github.com/sporkmonger/addressable/pull/492/files#diff-ebf4cdeb47084eeefe86a7b9e669d9b89adb907a1c5c19c2928233fabb2c4f86
They now test explicitely that:
1. unicode is normalized even if passed in as the wrong encoding
2. invalid UTF-8 chars are rejected (with an exception) like before.

Another approach could be to try to `encode` the incoming string to UTF-8 instead of just `force_encoding`, meaning a valid ISO string for example, would be converted to a different but valid UTF-8 string. But I think it's better to leave this behavior to the client as it could be too surprising. Also it could break some existing code usage were people already pass valid unicode string in the wrong encoding.

jarthod: @sporkmonger gentle ping :ping_pong: just to see if you agree, I'll start working on the other MR once this one is merged.

joseportillam (approved):

lib/addressable/template.rb
jarthod: @dentarg shall we keep waiting for Bob approval on this one? I don't want to be too pushy.

dentarg: @jarthod I can give him a nudge via email

Big disclaimer: I have not read the different RFCs as you two have, but I think your last comment makes sense.

jarthod: Ok thanks, there's no rush for me so better be safe and wait for @sporkmonger approval then, these RFCs are tricky so 2.5 pairs of eyes is always better :)

sporkmonger (approved):
Looks good to me!

sporkmonger:
Sorry about being slow to respond! I was out on vacation last week and the laptop stayed home.

jarthod:
@sporkmonger no problem, thank you!

lib/addressable/idna/native.rb
```diff
@@ -29,10 +29,6 @@ def self.punycode_decode(value)
        IDN::Punycode.decode(value.to_s)
      end
 
-    def self.unicode_normalize_kc(value)
-      IDN::Stringprep.nfkc_normalize(value.to_s)
-    end
```
dentarg, line 34: @jarthod Maybe we were a bit aggressive removing the method `Addressable::IDNA.unicode_normalize_kc` in a patch version. There might be code in the wild using these methods. Just happened to see [some code using `Addressable::IDNA.to_ascii`](https://github.com/discourse/discourse/pull/17840/files#diff-f98dab56f29862edb3b9003dde7224a286c23262bef7328698c4864c4ade1c6fR157).

Do you think it is possible to add it back? (In both native and pure versions, we don't know what people are using)

dentarg: Would be cool to mark it as deprecated (though I don't think this project as an established way of doing that, some new code would be needed)

jarthod: Why not, as you prefer. We can add it back easily, yes, with some caveats:
1. For `Pure` I don't think we should restore the old ruby implementation but use the new one (ruby's implementation) so it won't be 100% compatible with before but at least the method will exist.
2. Same thing for `Native`, if we restore the `IDN::Stringprep` one (meaning 100% identical to before the removal) the result will now be different than what addressable uses (ruby normalize). Also there's [no equivalent in `libidn2` AFAICS](https://libidn.gitlab.io/libidn2/reference/api-index-full.html). So we should also forward to ruby normalize I suppose.

---

For the deprecation, I didn't see any existing way either in the project. From a quick look I found this std ruby module: https://ruby-doc.org/stdlib-3.1.0/libdoc/rubygems/rdoc/Gem/Deprecate.html
Which seems to be [recommended](https://stackoverflow.com/a/23554720/304434) to do that (never used it). It prints this standard message we saw many times, would look like this (manually replacing the interpolations):
```
NOTE: Addressable::IDNA.unicode_normalize_kc is deprecated; use String#unicode_normalize(:nfkc) instead. It will be removed on or after 2023-04.
Addressable::IDNA.unicode_normalize_kc called from xxxxxxxx:42
```
The date is not optional unfortunately, but the message says "or after" so it's not a big deal we can leave "2023-04" so people know it's deprecated and will stop working at some point. This message allows to provide the replacement code which is good. Also it support the _standard_ `Gem::Deprecate.skip` option.

Alternatively we can [use `Kernel.warn` directly](https://stackoverflow.com/a/294114/304434), which even has a ["category" option which can take the value `:deprecated`](https://ruby-doc.org/3.2.1/Kernel.html#method-i-warn), though it's not clear what it changes and even Gem::Deprecate doesn't use it.

dentarg: Thanks yet again for an awesome comment @jarthod :) I agree with you that we should not restore the old behaviours, just bring back the methods so code doesn't break. I'm fine with either options for deprecating.

jarthod: → #504

Issue #247, opened 2016-11-03
Add test cases for IDNA2003 vs IDNA2008


sporkmonger:
Per http://www.unicode.org/reports/tr46/ – make sure Addressable supports IDNA2008 and add test cases on major edge cases between IDNA2003 and IDNA2008.

dentarg:
Looks like Addressable, when [used with libidn](https://github.com/sporkmonger/addressable/blob/addressable-2.8.1/lib/addressable/idna/native.rb) (via [`idn-ruby`](https://github.com/deepfryed/idn-ruby) [gem](https://rubygems.org/gems/idn-ruby)), follows IDNA 2003 (as stated by [libidn website](https://www.gnu.org/software/libidn/)) but when libidn (`idn-ruby` gem) isn't available, and we fall back to the ["pure" version](https://github.com/sporkmonger/addressable/blob/addressable-2.8.1/lib/addressable/idna/pure.rb), IDNA 2008 looks to be used.

```ruby
# native (libidn)
irb(main):001:0> Addressable::URI.parse("http://faß.de").normalize
=> #<Addressable::URI:0xf78 URI:http://fass.de/>

# pure
irb(main):001:0> Addressable::URI.parse("http://faß.de").normalize
=> #<Addressable::URI:0x13d8 URI:http://xn--fa-hia.de/>
```

Full repro:

```bash
arm64 $ docker run -it --rm ruby:3.1.3 bash
root@2a71c4abec4a:/# gem install addressable
Fetching public_suffix-5.0.1.gem
Fetching addressable-2.8.1.gem
Successfully installed public_suffix-5.0.1
Successfully installed addressable-2.8.1
2 gems installed
root@2a71c4abec4a:/# irb -raddressable/uri
irb(main):001:0> Addressable::URI.parse("http://faß.de").normalize
=> #<Addressable::URI:0x13d8 URI:http://xn--fa-hia.de/>
irb(main):002:0>
root@2a71c4abec4a:/# apt-get update && apt-get install -y libidn11-dev
Get:1 http://deb.debian.org/debian bullseye InRelease [116 kB]
Get:2 http://deb.debian.org/debian-security bullseye-security InRelease [48.4 kB]
Get:3 http://deb.debian.org/debian bullseye-updates InRelease [44.1 kB]
Get:4 http://deb.debian.org/debian bullseye/main arm64 Packages [8072 kB]
[...]

PR #496, opened 2023-03-22
libidn2 support for IDNA2008+UTS#46 (using ffi)

jarthod:
Following #247, here is the PR adding `libidn2` support through a small `ffi` wrapper (no dependencies other than `ffi`).

:heavy_check_mark:  Benchmark result (benchmark code in the PR), slightly slower than `libidn1` as it's doing more work but nothing to worry about, we're still above 100k iterations/sec which is good :rocket: :
```
# > ruby benchmark/idna.rb
# Rehearsal -------------------------------------------
# pure      5.914630   0.000000   5.914630 (  5.915326)
# libidn    0.518971   0.003672   0.522643 (  0.522676)
# libidn2   0.763936   0.000000   0.763936 (  0.763983)
# ---------------------------------- total: 7.201209sec

#               user     system      total        real
# pure      6.042877   0.000000   6.042877 (  6.043252)
# libidn    0.521668   0.000000   0.521668 (  0.521704)
# libidn2   0.764782   0.000000   0.764782 (  0.764863)
```

:heavy_check_mark: I also added a memory leak test to this benchmark code, to make sure I wasn't forgetting any manual memory free from C allocations. It's all good now and we can easily see by commenting one of the `idn2_free` lines the memory increasing so looking good on this side:
```
# Memory leak test for libidn2 (memory should stabilize quickly):
#  Memory: 117MB
#  Memory: 121MB
#  Memory: 121MB
#  Memory: 121MB
#  Memory: 121MB
#  Memory: 121MB
#  Memory: 121MB
#  Memory: 121MB
#  Memory: 121MB
#  Memory: 121MB
```

:heavy_check_mark: Specs are green locally (I just added 3 more to specify if the implementation is following IDNA2003 or IDNA2008+UTS#46) and on CI. I verified and there's no additional steps required to install libidn2 (already present by default in ubuntu and macos). I confirmed all the envs are running and passing libidn2 test (except Windows of course), we can see this with the increased number of test ran compared with master and the absence of the "Could not load native libidn2 implementation" line.

I saw maybe the `profile` job will have to be updated as it's currently running `idna_mode: [native, pure]` only. I'll have a look at this.
→ Edit: done, I've added the "native2" variant to the profile job and it [runs fine](https://github.com/sporkmonger/addressable/actions/runs/4495215047/jobs/7908613017). I confirmed locally that it's using libidn2 (and we can see `ffi` in the output). MemoryProfiler won't catch C-level memory leaks though (I tried).


addressable.gemspec
```diff
@@ -27,9 +27,11 @@ Gem::Specification.new do |s|
 
   if s.respond_to? :add_runtime_dependency then
     s.add_runtime_dependency(%q<public_suffix>.freeze, [">= 2.0.2", "< 6.0"])
+    s.add_runtime_dependency(%q<ffi>.freeze)
     s.add_development_dependency(%q<bundler>.freeze, [">= 1.0", "< 3.0"])
   else
     s.add_dependency(%q<public_suffix>.freeze, [">= 2.0.2", "< 6.0"])
+    s.add_dependency(%q<ffi>.freeze)
```
jarthod, line 34: This is optional, we could also decide to make `ffi` optional like the `idn-ruby` gem. Obviously open to discussion, my opinion is that:
- leaving `ffi` as an optional dependency could create surprises. For example I add another gem to my project which installs `ffi` and suddenly `addressable` behavior "changes" because it uses `libidn2`. Or the other way around, which could cause an unexpected performance drop.
- `ffi` being a quite common and well maintained ruby gem it's not a big risk
- also as opposed to `idn-ruby`, `ffi` provides `libidn2` which is the up-to-date implementation, hopefully matching pure ruby version perfeclty once #491 is addressed, but likely faster.

So IMO it's ok to add it as a runtime dependency so people benefit from it if available.

lib/addressable/idna/native.rb
```diff
@@ -16,19 +16,11 @@
...
 module Addressable
   module IDNA
-    def self.punycode_encode(value)
-      IDN::Punycode.encode(value.to_s)
-    end
-
-     def self.punycode_decode(value)
-       IDN::Punycode.decode(value.to_s)
-     end
-
```
jarthod, line 31: This is not used anywhere, not tested and not publicly available in the `pure` implementation (it's `private`). It's also not available in `libidn2`. So I think we can safely remove it for consistency. And even if somebody somewhere was actually using this method manually for some reason (and it was working because they used `libidn`), they can easily replace the one line call with the `IDN::Punycode` directly.

spec/addressable/idna_spec.rb
```diff
@@ -258,14 +258,18 @@
 end
 
 describe Addressable::IDNA, "when using the pure-Ruby implementation" do
-  before do
+  before :all do
```
jarthod, line 261: Small optimisation to avoid wasting useless CPU cycles reloading the whole files before EACH test. It's more than 10 times faster (but was already very fast so not a big deal)

addressable.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
...
+  s.files = ["CHANGELOG.md".freeze, "Gemfile".freeze, "LICENSE.txt".freeze, "README.md".freeze, "Rakefile".freeze, "data/unicode.data".freeze, "lib/addressable".freeze, "lib/addressable.rb".freeze, "lib/addressable/idna".freeze, "lib/addressable/idna.rb".freeze, "lib/addressable/idna/native.rb".freeze, "lib/addressable/idna/native2.rb".freeze, "lib/addressable/idna/pure.rb".freeze, "lib/addressable/template.rb".freeze, "lib/addressable/uri.rb".freeze, "lib/addressable/version.rb".freeze, "spec/addressable".freeze, "spec/addressable/idna_spec.rb".freeze, "spec/addressable/net_http_compat_spec.rb".freeze, "spec/addressable/security_spec.rb".freeze, "spec/addressable/template_spec.rb".freeze, "spec/addressable/uri_spec.rb".freeze, "spec/spec_helper.rb".freeze, "tasks/clobber.rake".freeze, "tasks/gem.rake".freeze, "tasks/git.rake".freeze, "tasks/metrics.rake".freeze, "tasks/profile.rake".freeze, "tasks/rspec.rake".freeze, "tasks/yard.rake".freeze]
```
jarthod, line 16: I was very inpired so I choose `native2` as file name :rofl:, we could also rename both files to `libidn1`/`libidn` and `libidn2` for example, let me know if you prefer.

tasks/profile.rake
```diff
@@ -41,13 +41,16 @@ namespace :profile do
...
+      Addressable.send(:remove_const, :IDNA)
+      load "addressable/idna/native.rb"
     end
 
     start_at = Time.now.to_f
     report = MemoryProfiler.report do
       30_000.times do
         Addressable::URI.parse(
-          "http://google.com/stuff/../?with_lots=of&params=asdff#!stuff"
+          "http://fiᆵリ宠퐱卄.com/stuff/../?with_lots=of&params=asdff#!stuff"
```
jarthod, line 53: Changed the hostname to include the IDNA conversion in this profiling.

jarthod:
:heavy_check_mark: As usual I also verified this new version against my 150k URL at updown.io, no differences after normalization between master (using pure implementation) and this branch (using native2). Except that it took only 5.6 seconds now (versus 12s with pure) to normalize the 150k URLs.

benchmark/idna.rb
```diff
@@ -0,0 +1,65 @@
...
+
+puts "\nMemory leak test for libidn2 (memory should stabilize quickly):"
+load "lib/addressable/idna/native2.rb"
+GC.disable # Only run GC when manually called
+10.times do
+  N.times { Addressable::IDNA.to_unicode(Addressable::IDNA.to_ascii(value)) }
+  GC.start # Run a major GC
+  pid, size = `ps ax -o pid,rss | grep -E "^[[:space:]]*#{$$}"`.strip.split.map(&:to_i)
+  puts " Memory: #{size/1024}MB" # show process memory
+end
```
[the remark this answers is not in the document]
dentarg: That sure would be sweet 🤩  Here or in a separate PR?

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         ruby: [2.7]
-        idna_mode: [native, pure]
+        idna_mode: [native2, native, pure]
```
dentarg, line 16: Don't we also need to install Libidn2 in CI?

addressable.gemspec
dentarg: Yeah it is tempting to make these names more clear... as in my mind both "pure" and "native" could be names for the (pure) Ruby implementation. :) I guess we would need to keep `lib/addressable/idna/native.rb` around though (requiring `libidn1`) to be backwards compatible?

addressable.gemspec
dentarg: There currently is a [rake task](https://github.com/sporkmonger/addressable/blob/e91b64e1ab36dccacd8e339a84e89684ed5bbb3e/tasks/gem.rake#L5-L50) to that the gemspec is generated from that needs to be update in order to preserve these changes. (I'm not a big fan of this rake task, any new contributor doesn't know about it and I even forget it myself sometimes, and it specifies things you don't need to specify in the gemspec, and the [file listing is strange](https://github.com/sporkmonger/addressable/pull/497) (should change it to use git)).

I think you're right we should depend on `ffi` to make it as simple as possible to use `libidn2`... with that said, how are we going to roll out this change? In a minor version? In a major version? Should it be opt-in or opt-out to use `libidn2`?

lib/addressable/idna.rb
```diff
@@ -16,11 +16,15 @@
...
-end
+  # libidn2 or the ffi gem was not available, fall back on libidn1
+  begin
+    require "addressable/idna/native"
+  rescue LoadError
+    # libidn or the idn gem was not available, fall back on a pure-Ruby
+    # implementation...
+    require "addressable/idna/pure"
+  end
+end
```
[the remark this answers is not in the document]
dentarg: I like your suggestions. This has been (and still is) a bit of a gotcha (the tests have `IDNA_MODE` to control it). Sounds like the improvements needs to be introduced in a major version though?

lib/addressable/idna/native.rb
dentarg: Yes, agree.

lib/addressable/idna/native2.rb
```diff
@@ -0,0 +1,58 @@
...
+      raise "libidn2 failed to convert \"#{value}\" to ascii (#{idn2_strerror(res)})" if res != 0
+      result = pointer.read_pointer.read_string
+      idn2_free(pointer.read_pointer)
+      result
+    end
+
+    def self.to_unicode(value)
+      pointer = FFI::MemoryPointer.new(:pointer)
+      res = idn2_to_unicode_8z8z(value, pointer, IDN2_NONTRANSITIONAL)
+      return value if res != 0
```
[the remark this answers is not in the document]
dentarg: Something to change in a major version bump?

.github/workflows/test.yml
dentarg: Apparently not, looks like [`libidn2`](https://packages.ubuntu.com/focal/libidn2-0) is probably a dependency to [many packages installed on the runner images](https://github.com/actions/runner-images/blob/833ffa3aab68f80abf0b0e987e212f43da58f925/images/linux/Ubuntu2004-Readme.md#installed-apt-packages), e.g. [libcurl4](https://packages.ubuntu.com/bionic/libcurl4) or [libc6](https://packages.ubuntu.com/focal/libc6)

addressable.gemspec
jarthod: Aha indeed I missed this one ^^ this is the older way of doing gems so probably just legacy.
Also not a big fan of this way of doing, but let's not mix such independant change in this branch for the moment.

I addressed this in 0a6f09194a66f07e0b095ce87c844848547d497a by:
- Adding the `ffi` dependency in the rake task
- Updating the file regexp to remove the useless folders and add back the missing files
- Run the task and compared the output with main to make sure the only difference is `native2.rb` and `ffi` dependency. (aside from date and rubygems version ^^)
So now running `gem:release` should not cause any unwanted change.

About the behavior I believe it should be opt-out like it's currently the case with libidn, if it's available we just use it because it's supposed to be the "best" implementation. Fallback to `libidn` if available, and `pure` otherwise.
And if somebody wants to force one mode explicitely, I would say they can use something like `Addressable::IDNA = Addressable::IDNA::Pure` as suggested in my other comment (I'll add this to the readme if we do that)

About the version bump that is a good question, even though it's not breaking the API, it would still change the internal behavior a bit (especially IDNA2008+UTS46), some class names, etc. So I would say it could deserve a major version bump. But in that case, we should probably bundle #491 in the same major version bump, this way we ship one major update which brings IDNA2008+UTS46 to BOTH implementations at the same time, with the ability to still use the older IDNA2003 using `libidn` for people who needs. WDYT?

addressable.gemspec
jarthod: Definitely agree, I find "native" naming a bit confusing as we often say "native ruby" for something with no C extensions ^^
I'll do that then, and yes it's true we can keep the old file with just a require for compatibility (and/or maybe a deprecation warning)

lib/addressable/idna.rb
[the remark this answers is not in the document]
jarthod: This wasn't as easy as I hoped because:
- `Addressable::IDNA = Addressable::IDNA::Native2` is not possible, if we re-assign `Addressable::IDNA` then `Addressable::IDNA::Native2` is no longer accessible because it was INSIDE the previous module we just shadowed by re-assigning. We can workaround this by changing the name of one or the other, but anyway:
- Re-assigning `Addressable::IDNA` is also not something Ruby likes simply because it's considered a CONSTANT so gives warnings when we change it (not a big deal but we can't recommend this as a way to select the backend if it generates warnings)

In the end I implented it another way in 62a136576446912fddde671ba41088df10ee9104:
- I set the backend in a class variable instead of a constant e.g: `Addressable::IDNA.backend = Addressable::IDNA::Libidn2`
- We can easily check the backend with `Addressable::IDNA.backend.name` (used it in profile:memory as an example)
- In order not to break compatibility with people who may be using `Addressable::IDNA.to_ascii` and `Addressable::IDNA.to_unicode`, I defined them as simple method in the IDNA module (delegating to the backend).
- I also changed the module/file names as discussed below at the same time (`libidn2`, `libidn1` and `pure`)

This has a couple side benefits:
- The common code like (deprecated) `unicode_normalize_kc` method can now be inside this `Addressable::IDNA` module instead of being duplicated and cluturing the backend modules.
- The public API is much clearer now (`to_ascii` and `to_unicode`) and other private or public method and constants in the backend are NOT accessible from `Addressable::IDNA`, reducing this kind of problem in the future. (people who want to hack can still call them from the backend directly, but in that case it's their problem).

Some downsides I see:
- The git history is a bit lost for these files, it'll show this comit because I namespaces the module. As the `pure` one which is the biggest is also gonna change a lot in #491, maybe not a big deal
- We can't really remove `IDNA_MODE` because even though it's simpler to switch in Ruby, we still need a way for the text matrix to communicate that choice to the code.

I updated the readme with some instructions and doc about this if you want a clearer picture.
Let me know what you think, and if you are not comfortable with this change we can also forget about it, no big deal.

addressable.gemspec
dentarg: Everything above sounds good to me :) 👍

dentarg: One alternative idea is adding this (`libidn2` support) in a minor version as opt-in (and make it opt-out (with #491) in a future major version), the pitch being it would be possible for users to prepare for the major version by upgrading to the minor version and switching backend back and forth. Does it makes sense or would it be too much work?

benchmark/idna.rb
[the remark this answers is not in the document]
jarthod: Let's do this then, I just added the `profile:idna_memory_leak` task in 421cdcaa019b7ecc30f944d8c0131770b71d8829

It looks like this with a simulated memory leak (by commenting one of the `idn2_free` lines):
github com_sporkmonger_addressable_actions_runs_4671950392_jobs_8273572062 (1)

I tried to re-use `profile:memory` but I can't unfortunately because `MemoryProfiler` does not track the total memory at all (so it can't detect any kind of C memory leaks). I also tried to mix both tests together, but the `GC.start` needed for the memory leak test actually breaks `MemoryProfiler` because it needs to keep all objects for analysis ^^

lib/addressable/idna/native.rb
```diff
@@ -1,66 +1,4 @@
...
-          ''
-        end
-      end.join('.')
-    end
-  end
-end
+# Deprecated, for backward compatibility only
+require "addressable/idna/libidn1"
+Addressable::IDNA.backend = Addressable::IDNA::Libidn1
+warn "NOTE: loading 'addressable/idna/native' is deprecated; use 'addressable/idna/libidn1' instead and set `Addressable::IDNA.backend = Addressable::IDNA::Libidn1` to force libidn1."
```
dentarg, line 4: Can you fix the "No newline at end of file"? :) There's another one for spec/addressable/idna_spec.rb too

lib/addressable/idna/pure.rb
```diff
@@ -97,7 +97,7 @@ def self.to_unicode(input)
         if part =~ /^#{ACE_PREFIX}(.+)/
           begin
             punycode_decode(part[/^#{ACE_PREFIX}(.+)/, 1])
-          rescue Addressable::IDNA::PunycodeBadInput
+          rescue Addressable::IDNA::Pure::PunycodeBadInput
```
dentarg, line 100: Oh, it is possible to not change the namespace of these errors? If we are not going for a major version now. I know at least one library (authored by me in the past, hehe) that rely on these names (well, [only one so far](https://github.com/twingly/twingly-url/issues/112)): https://github.com/twingly/twingly-url/blob/e20f5fce077d93e89ef8520961be453c90cfec8c/lib/twingly/url.rb#L21

dentarg: I guess having these exceptions defined under `Addressable::IDNA::Pure` is more correct though, if they can only be raised from code in `Addressable::IDNA::Pure`? 

Can `Addressable::IDNA::Libidn2` raise any exceptions? Looks like `to_ascii` can raise `RuntimeError`. `to_unicode` being discussed at https://github.com/sporkmonger/addressable/pull/496#discussion_r1145256087

benchmark/idna.rb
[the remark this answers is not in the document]
dentarg: Really nice

lib/addressable/idna.rb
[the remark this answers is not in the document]
dentarg: I like the way you took. Greate updates to the README.

I'm not too worried about the git history when we are making motivated changes, but there's always the [`.git-blame-ignore-revs`](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) option to mitigate impact, but I think we are fine in this case.

README.md
```diff
@@ -94,15 +94,45 @@ template.extract(uri)
...
+- A pure ruby implementation (slower), [almost](https://github.com/sporkmonger/addressable/issues/491) supporting IDNA2008.
+- A `libidn2` wrapper (if `libidn2` is installed), supporting IDNA2008+UTS#46.
+
+Note: in the future major version, `libidn2` will become the default.
+
+To install `libidn2`:
+
+```console
+$ sudo apt-get install libidn2-dev # Debian/Ubuntu
+$ brew install libidn # OS X
```
dentarg, line 110:
```suggestion
$ brew install libidn2 # OS X
```

dentarg (commented):
A few comments, need to take this for a spin too when time permits :)

lib/addressable/idna/native2.rb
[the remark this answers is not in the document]
jarthod: This one and 2 others would be failing if I raise an error here:
```
  1) Addressable::IDNA when using the libidn2 native implementation (ffi) it should behave like converting from ASCII to unicode should convert 'AcinusFallumTrompetumNullunCreditumVisumEstAtCuadLongumEtCefallum.com' correctly
     Failure/Error: raise "libidn2 failed to convert \"#{value}\" to unicode (#{idn2_strerror(res)})" if res != 0
     
     RuntimeError:
       libidn2 failed to convert "AcinusFallumTrompetumNullunCreditumVisumEstAtCuadLongumEtCefallum.com" to unicode (domain label longer than 63 characters)
     Shared Example Group: "converting from ASCII to unicode" called from ./spec/addressable/idna_spec.rb:321
     # ./lib/addressable/idna/libidn2.rb:52:in `to_unicode'
     # ./lib/addressable/idna.rb:30:in `to_unicode'
     # ./spec/addressable/idna_spec.rb:160:in `block (2 levels) in <top (required)>'

  2) Addressable::IDNA when using the libidn2 native implementation (ffi) it should behave like converting from ASCII to unicode should return the identity conversion when punycode decode fails
     Failure/Error: raise "libidn2 failed to convert \"#{value}\" to unicode (#{idn2_strerror(res)})" if res != 0
     
     RuntimeError:
       libidn2 failed to convert "xn--zckp1cyg1.sblo.jp" to unicode (string contains invalid punycode data)
     Shared Example Group: "converting from ASCII to unicode" called from ./spec/addressable/idna_spec.rb:321
     # ./lib/addressable/idna/libidn2.rb:52:in `to_unicode'
     # ./lib/addressable/idna.rb:30:in `to_unicode'
     # ./spec/addressable/idna_spec.rb:164:in `block (2 levels) in <top (required)>'

  3) Addressable::IDNA when using the libidn2 native implementation (ffi) it should behave like converting from ASCII to unicode should return the identity conversion when the ACE prefix has no suffix
     Failure/Error: raise "libidn2 failed to convert \"#{value}\" to unicode (#{idn2_strerror(res)})" if res != 0
     
     RuntimeError:
       libidn2 failed to convert "xn--...-" to unicode (string contains invalid punycode data)
     Shared Example Group: "converting from ASCII to unicode" called from ./spec/addressable/idna_spec.rb:321
     # ./lib/addressable/idna/libidn2.rb:52:in `to_unicode'
     # ./lib/addressable/idna.rb:30:in `to_unicode'
     # ./spec/addressable/idna_spec.rb:169:in `block (2 levels) in <top (required)>'
```
The last two are invalid punycode and the first one is invalid DNS length (https://datatracker.ietf.org/doc/html/rfc1034#section-3.1).

`libidn1` also raise in the first case but we have this workaround to explicitely allow for > 63 bytes labels:
```ruby
    def self.to_ascii(value)
      value.to_s.split('.', -1).map do |segment|
        if segment.size > 0 && segment.size < 64
          IDN::Idna.toASCII(segment, IDN::Idna::ALLOW_UNASSIGNED)
        elsif segment.size >= 64
          segment
        else
          ''
        end
      end.join('.')
    end
```
Looks like this was made in c73810f4823c4dfac1f0568073623eb7922f526d to make it more consistent with `pure`. Didn't see any issue attached though. 

So I suppose if we make `libidn2` stricter, which means basically:
```ruby
raise "libidn2 failed to convert \"#{value}\" to unicode (#{idn2_strerror(res)})" if res != 0
```
instead of
```ruby
return value if res != 0
```
We would need to remove these workarounds and make all implementations rejects these domains in the same way.
Which does sound like the way to go IMO but of course could break some use-cases for people who need to handle such "slightly" invalid domains.

lib/addressable/idna/pure.rb
jarthod: Good point, It could make sense to move these exception back up in `IDNA` and if possible re-use them in `libidn1` and `libidn2` then. Though at the moment I made a very simple error handling so only the message changes. Distinguishing between different types of errors in `libidn2` would require keeping a mapping of native error code ([example from ruby-idna](https://github.com/ogom/ruby-idna/blob/master/lib/idna/error.rb)), I'd rather avoid that if possible.

Here is what I suggest: let's add a generic `Addressable::IDNA::Error` (which I will use in `libidn2` instead of `RuntimeError`, it's actually something I wanted to do but forgot) and move back the pure exceptions to IDNA to keep compatibility but make them inherit from `Addressable::IDNA::Error`. This way the old exception don't change, and people can rescue `Addressable::IDNA::Error` if they want all of them (this one is future-proof).

I just did that in 84fdb89ef413f60b219f89c9910604247d3b84f0

lib/addressable/idna/native2.rb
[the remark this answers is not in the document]
dentarg: Yeah, I'm thinking c73810f4823c4dfac1f0568073623eb7922f526d was the wrong call, it should have been the pure version that changed? 

I'm down with making too long names invalid, question is, does it need to happen in a major version? 🤔 c73810f4823c4dfac1f0568073623eb7922f526d happened in 2.5.0

* [this wrapper wants it](https://github.com/twingly/twingly-url/issues/141)
* Chrome doesn't try to visit `AcinusFallumTrompetumNullunCreditumVisumEstAtCuadLongumEtCefallum.com`, it makes a search engine search for me
* ```
  $ curl -s -v AcinusFallumTrompetumNullunCreditumVisumEstAtCuadLongumEtCefallum.com
  * Could not resolve host: AcinusFallumTrompetumNullunCreditumVisumEstAtCuadLongumEtCefallum.com
  * Closing connection 0
  ```

The other two are harder, I don't know what's correct. They come from 295d7baec1281976434e9061b8065979b3f38c83 (`The spec says `toUnicode` returns the identity conversion on bad decodes.`) and cf8cc7d0b9c8d2df9caeff90a063864c3e68d940 (`Correctly handle the case where an ACE prefix has no suffix.`) – probably not good changing these in a non-major version

What if we introduce a "strict mode"? `Addressable::IDNA.strict_mode` (the user can change it) that's `false` in the next non-major version, keeping current behaviour, and `true` in the next major version meaning we raise these errors.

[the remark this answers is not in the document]
jarthod: > Yeah, I'm thinking https://github.com/sporkmonger/addressable/commit/c73810f4823c4dfac1f0568073623eb7922f526d was the wrong call, it should have been the pure version that changed?

That is what I think too.

Why not for the `strict_mode` (which would basically–if falsy–hide the errors and return the original name unchanged), this could even be implemented at `Addressable::IDNA` level so it supports all backends. Usually I would say this can be done on the client side so it's client responsibility, but here we're talking about one internal component of Addressable and people may want to normalize a URL even if the hostname is invalid for example. So I agree to add an option for that.

I'll try to do that.

addressable.gemspec
jarthod: Do have any other feedback on this part?

dentarg: I think what you did was great :)

lib/addressable/idna/native2.rb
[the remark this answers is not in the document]
jarthod: Ok so after giving this a try, it's not as easy as expected of course. If I add a `strict_mode` and make things right for `libidn2` (so let the native exceptions go when the string is invalid), then we get consistency problems with the other backends of course:
1. `libidn1` (even after removing the change from c73810f4823c4dfac1f0568073623eb7922f526d) does **not** raise when the label is > 63 bytes and we're converting from unicode to ascii (only the other way around). It's not a big deal (it's less strict than libidn2) but it's inconsistent so I can't just spec this behavior on all backends and would need to spec this only for `libidn2`, or we could add this exception ourselves in ruby.
2. `libidn1` does **not** raise when the punnycode is invalid like `libidn2` (I tried other flags but none of them make it raise in these two cases). Also not a big deal IMO, it's just that the strict mode will be more strict with libidn2 than libidn1. I don't see any easy way to fix that in ruby.
3.  The current `pure` implementation (even after removing the change from cf8cc7d0b9c8d2df9caeff90a063864c3e68d940) does not raise (it returns a wrong result instead), I can make it raise with small changes. For the segment length no check is made at all so we would need to add them too here if we want to be consistent, which does not look easy (it skips conversion anyway for ascii strings so we can't just add a length check in the segment loop, etc...).

So quick question here: do you think it's ok if the `strict_mode` has a different strictness depending on the backend? or shall we drop this idea and keep only the more tolerating mode for all?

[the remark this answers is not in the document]
dentarg: What about this idea: as "strict mode" is something new, it could be a thing only for libidn2 

Does that make sense?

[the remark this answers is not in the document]
jarthod: Yep, I just added the strict mode in a5e87c6824ff927adc05db4d21103fccb4ce3277.
It applies mostly to `libidn2` at the moment, though I implemented it at global level and still removed the change in libidn1 to simplify the code and make it slightly stricter (when the option is true of course).

Let me know what you think :)

README.md
```diff
@@ -94,15 +94,46 @@ template.extract(uri)
...
+
+To install `libidn1` and the `idn` gem (also add it to your Gemfile):
 
 ```console
 $ sudo apt-get install libidn11-dev # Debian/Ubuntu
 $ brew install libidn # OS X
 $ gem install idn-ruby
 ```
 
+Optionally you can turn on the strict mode which will raise exceptions in case of invalid hostname during IDNA conversion. The default (`false`) silently ignores them and keeps the hostname unchanged. The strictness will depend on the backend used, libidn2 is stricted than libidn1 for example.
```
dentarg, line 121:
```suggestion
Optionally you can turn on the strict mode which will raise exceptions in case of invalid hostname during IDNA conversion. The default (`false`) silently ignores them and keeps the hostname unchanged. The strictness will depend on the backend used, libidn2 is stricter than libidn1 for example.
```

lib/addressable/idna/native2.rb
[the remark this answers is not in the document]
dentarg: Looks good to me :) 

Need to take the time to look at/play with this PR as a whole, not sure exactly when that will be though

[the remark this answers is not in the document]
jarthod: Ok thanks, I've run again my 150k URL through the new "strict-mode" conversion, and it found a couple invalid ones which I double checked and are really invalid (none of them ever worked) so that is good for me. This would actually be even better for me to be able to reject those before creation so I'll probably use `strict_mode=true`.

It raised one question for me though, at the moment I rescued `Addressable::URI::InvalidURIError` to detect invalid URLs (which is theoretically not enough as we could also have `Addressable::IDNA::PunycodeBadInput`, `Addressable::IDNA::PunycodeBigOutput`, `Addressable::IDNA::PunycodeOverflow` with the Pure implementation I'm using but I've never got those). And now there can also be `Addressable::IDNA::Error` with `libidn2`.

That is fine for me and I can start rescuing `Addressable::IDNA::Error` too (which is parent to the others Pure-specific exeptions now so good enough) but I'm thinking about other users who may not be aware of this, it could be a bit surprising. You already had this problem with https://github.com/twingly/twingly-url/issues/112 so what would be your preference? I am thinking about these options:
1. Keep it like this, and people can rescue `Addressable::IDNA::Error` for all IDNA related errors (that would simplify a bit your twingly fix by the way and make it more future-proof)
2. Make Addressable::URI rescue the IDNA exceptions, and re-raise it as `Addressable::URI::InvalidURIError`, this way the API has more limited boundaries which means only one exception to rescue for users (unless they are using the IDNA module directly of course).
3. Alternativaly I just thought about this little hack which sounds easy and nice yet does not _feels_ right :upside_down_face: : make `Addressable::IDNA::Error` inherit from `Addressable::URI::InvalidURIError` instead of `StandardError`. This way people who do `rescue Addressable::URI::InvalidURIError` are actually already covered and we don't have to rescue/re-raise or change anything else. It kinda make sense as if the IDNA host is invalid it definitely means the URL is invalid, and I can't find any counter-example where this would cause trouble.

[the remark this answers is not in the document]
dentarg: I think 3. makes sense, it should give users the most flexibility 

Can we make that change now in the next minor version this would go into, or could it be considered a breaking change? I'm leaning towards it not being a breaking change. WDYT?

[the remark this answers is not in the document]
jarthod: I just tried it in ccbcb9e419de67ed17630908f229bf03e41acecf , it makes sense to me too though I'm not a big fan of leaking some of "URI" into "IDNA" (IDNA being supposedly a dependency to URI), still feels wrong but works fine :)

I think this can be made in the same minor yes, I don't see anything breaking in here as we kept the default being `safe_mode=false` and the existing exceptions haven't changed class (just parent).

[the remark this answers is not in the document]
dentarg: Alternative to \3., with the same outcome but without changing the inheritance would be to use `#extend`: https://github.com/twingly/twingly-url/pull/106#issuecomment-258642114

Would still be leaking but maybe a little bit cleaner? :)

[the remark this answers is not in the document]
jarthod: Interesting suggestion, I do like the fact that this solution being run-time, we can use this class even if it's not defined yet. Unfortunately it doesn't work with classes, only modules: `wrong argument type Class (expected Module)` if I do `error.extend(Addressable::URI::InvalidURIError)`. In your example `Twingly::URL::Error` is actually a module. And if we need to change `Addressable::URI::InvalidURIError` to be a module this would complexify the rest :/

Looking again at `uri.rb` I see `IDNA` is used only twice (in `normalized_host` for `to_ascii` and in `display_uri` for `to_unicode`), so option 2 which is to re-wrap the error here doesn't sound too complicated either.

I just gave this option a try in 9eb3910a6376fa2c36ad0a9ab04285f49905a761 and I think I actually prefer this one. No hierarchy issue here, every module/class only deal with its own exceptions. The wrapped exception are properly identified and the `cause` attribute contains the previous exception (with specs for that just in case), this means backtrace and history is complete for bug tracker. People doing `rescue Addressable::URI::InvalidURIError` are covered the same way.

I also added by the way specs for the case of invalid IDNA hostname at URI level this time (I couldn't find any at the moment). 3 of them for the Pure implementation have been marked pending because they are returning garbage at the moment (implementation makes up unicode characters from invalid input).

And I also fixed the `libidn1` exception handling, which was still letting `IDN::Idna::IdnaError` exception up so not handled properly (I missed it earlier because there was no spec on this case), now it's raising `Addressable::IDNA::Error` like the other backends (+spec)

Sorry for the long back and forth, let me know what you think about this one ;)

lib/addressable/uri.rb
```diff
@@ -1140,6 +1140,8 @@ def normalized_host
       # All normalized values should be UTF-8
       force_utf8_encoding_if_needed(@normalized_host)
       @normalized_host
+    rescue IDNA::Error => e
+      raise InvalidURIError.new(e)
```
dentarg, line 1144: This is just an implicit `e.to_s` call, this could be clearer code
```suggestion
      raise InvalidURIError.new(e.message)
```

lib/addressable/idna/native2.rb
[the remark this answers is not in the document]
dentarg: Oh sorry for that, I did know about the `module` thing but forgot before I posted, oh well

Yes, I like https://github.com/sporkmonger/addressable/commit/9eb3910a6376fa2c36ad0a9ab04285f49905a761 too :) Thanks for extending the spec coverage

lib/addressable/uri.rb
jarthod: Sure, I tend to like the implicit look as: "Look I'm just wrapping this exception" but no problem I just updated them for the explicit look in 7d75d6c0c27be22411870709cab3db5ae438eadc

skryukov:
Hey! I really love interchangeable IDNA backends introduced in this PR. Would it be possible to merge that part? I've tested this PR with my [pure-Ruby implementation](https://github.com/skryukov/uri-idna) of IDNA2008/UTS46, and it integrates seamlessly:

```ruby
# frozen_string_literal: true

require "uri/idna"

module Addressable
  module IDNA
    module UTS46
      def self.to_ascii(value)
        URI::IDNA.to_ascii(value)
      end

      def self.to_unicode(value)
        URI::IDNA.to_unicode(value)
      end
    end
  end
end
```

The IDNA backend concept would allow me to easily package the Addressable backend within the gem.

Overall the `uri-idna` gem is 2-4 times slower than `Addressable::IDNA::Pure`, but I would prefer using that instead of the `ffi` version tbh :sweat_smile:

dentarg:
Cool. Maybe possible. I really want to find some time to properly play with this myself and merge it. It is really awesome work by @jarthod here. And I'm also following your repo @skryukov :) And https://github.com/ruby/uri/issues/76 hehe.

jarthod:
That would be great indeed, and excellent work on the pure Ruby implementation @skryukov :clap: (I'm also following https://github.com/ruby/uri/issues/76 ^^). @dentarg let me know if there is anything I can do to make it easier for you :bow:

lib/addressable/idna/native2.rb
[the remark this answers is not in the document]
sporkmonger: Necro-ing a little bit, but wanted to weigh in on the start of the conversation. I think it's important to offer a mechanism that's permissive in what it accepts. It's literally the reason I wrote Addressable in the first place, because the standard library doesn't take this approach to parsing and I couldn't parse URIs that were openable in a browser, leading to surprise from end users. There are often cases where failing with an exception will mean that there's no graceful way to get partial information. For instance, something might be very wrong with the encoding in the hostname, but if the library's user was only trying to retrieve the path value, the invalid URI exception is rather obstructive to that goal.

On the other hand, you're absolutely right that there are cases where the opposite is preferred and strict parsing is preferable. My view is these should simply be handled by different methods rather than changing the behavior for the whole library in a major version rev.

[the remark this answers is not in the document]
jarthod: @sporkmonger thanks for your feedback.

I understand and agree, the concensus that @dentarg and I reached further down this discussion was to introduce the `Addressable::IDNA.strict_mode = true` option (**default to false**) so that people can choose if they want stricter parsing but otherwise it's lax as before.

Of course if you prefer different methods instead of an option, we can probably do that instead. Though if there's different methods in IDNA modules for both behaviors, we would also need to mirror that in the URI module because the methods people call usually are here. I haven't checked the whole public API recently but I'm concerned this may create a lot of new methods. I also thought about people using gems _which depends on adressable_ : if we're using different methods, the end-user won't be able to change the behavior. I can have a deeper look if you want me to :+1:

jarthod:
For the record, I just rebased this branch on main, in case you want to go back to it as some point.
I'm still using it in production with 150k different URLs for the past 3 years.

PR #497, opened 2023-04-01
Correct files list in gemspec

dentarg:
The result after running `rake gem:release`.

Removed:

    ["benchmark", "data", "gemfiles", "lib", "lib/addressable", "lib/addressable/idna", "spec", "spec/addressable", "tasks"]


dentarg:
These jumped back with `b e rake gem:gemspec` I just did for https://github.com/sporkmonger/addressable/pull/503 (but I choose to not include that in the commit)

Issue #498, opened 2023-04-03
 undefined method `to_str' for :id:Symbol (NoMethodError) in 2.8.2


johrstrom:
Hi, I'm getting this error when I've updated to `2.8.2` from `2.8.1`.   Here's a very simple test case in `irb` that I've confirmed works in `2.8.1`. 

```irb
irb(main):004:0> uri = Addressable::Template.new("https://example/some/api/{?query*}")
=> #<Addressable::Template:0x104 PATTERN:https://example/some/api/{?query*}>
irb(main):005:0> uri.expand({'query' => {id: 'abc123' } })
/home/jeff/.gems/ruby-3.0.3/gems/addressable-2.8.2/lib/addressable/template.rb:900:in `normalize_value': undefined method `to_str' for :id:Symbol (NoMethodError)
Did you mean?  to_se-2.8.2.gem
               to_symx-5.0.1.gem
Successffrom /home/jeff/.gems/ruby-3.0.3/gems/addressable-2.8.2/lib/addressable/template.rb:908:in `block in normalize_value'
Successffrom /home/jeff/.gems/ruby-3.0.3/gems/addressable-2.8.2/lib/addressable/template.rb:907:in `each'
Parsing from /home/jeff/.gems/ruby-3.0.3/gems/addressable-2.8.2/lib/addressable/template.rb:907:in `inject'
Installifrom /home/jeff/.gems/ruby-3.0.3/gems/addressable-2.8.2/lib/addressable/template.rb:907:in `normalize_value'
Parsing from /home/jeff/.gems/ruby-3.0.3/gems/addressable-2.8.2/lib/addressable/template.rb:774:in `block in transform_capture'
Installifrom /home/jeff/.gems/ruby-3.0.3/gems/addressable-2.8.2/lib/addressable/template.rb:756:in `each'
Done insfrom /home/jeff/.gems/ruby-3.0.3/gems/addressable-2.8.2/lib/addressable/template.rb:756:in `inject'
2 gems ifrom /home/jeff/.gems/ruby-3.0.3/gems/addressable-2.8.2/lib/addressable/template.rb:756:in `transform_capture'
[dashboafrom /home/jeff/.gems/ruby-3.0.3/gems/addressable-2.8.2/lib/addressable/template.rb:595:in `block in expand'
irb(mainfrom /home/jeff/.gems/ruby-3.0.3/gems/addressable-2.8.2/lib/addressable/template.rb:594:in `gsub!'
[...]

PR #499, opened 2023-04-04
Fix template expand level 2 hash support for non-string objects

jarthod:
Here's the fix for #498

### DIFF
--- a/lib/addressable/template.rb
+++ b/lib/addressable/template.rb
@@ -896,19 +896,16 @@ module Addressable
     #
     # @return [Hash, Array, String] The normalized values
     def normalize_value(value)
-      unless value.is_a?(Hash)
-        value = value.respond_to?(:to_ary) ? value.to_ary : value.to_str
-      end
-
       # Handle unicode normalization
-      if value.kind_of?(Array)
-        value.map! { |val| normalize_value(val) }
+      if value.respond_to?(:to_ary)
+        value.to_ary.map! { |val| normalize_value(val) }
       elsif value.kind_of?(Hash)
         value = value.inject({}) { |acc, (k, v)|
           acc[normalize_value(k)] = normalize_value(v)
           acc
         }
       else
+        value = value.to_s if !value.kind_of?(String)
         if value.encoding != Encoding::UTF_8
           value = value.dup.force_encoding(Encoding::UTF_8)
         end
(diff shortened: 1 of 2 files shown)


lib/addressable/template.rb
```diff
@@ -896,19 +896,16 @@ def join_values(operator, return_value)
...
     def normalize_value(value)
-      unless value.is_a?(Hash)
-        value = value.respond_to?(:to_ary) ? value.to_ary : value.to_str
-      end
-
       # Handle unicode normalization
-      if value.kind_of?(Array)
-        value.map! { |val| normalize_value(val) }
+      if value.respond_to?(:to_ary)
+        value.to_ary.map! { |val| normalize_value(val) }
```
jarthod, line 901: Here I just simplified a bit the code, treating all array-like object in the same place instead of having the first like (900) to convert first.

lib/addressable/template.rb
```diff
@@ -896,19 +896,16 @@ def join_values(operator, return_value)
...
       elsif value.kind_of?(Hash)
         value = value.inject({}) { |acc, (k, v)|
           acc[normalize_value(k)] = normalize_value(v)
           acc
         }
       else
+        value = value.to_s if !value.kind_of?(String)
```
jarthod, line 908: And this is the fix basically, restoring the `to_s` behavior which I accidentally changed for `to_str` when I modified this method to make it recursive.

spec/addressable/template_spec.rb
```diff
@@ -209,7 +209,7 @@
       :path => "/foo/bar",
       :semi => ";",
       :list => %w(red green blue),
-      :keys => {"semi" => ';', "dot" => '.', "comma" => ','}
+      :keys => {"semi" => ';', "dot" => '.', :comma => ','}
```
jarthod, line 212: I noticed this specs are weird to write (you need to write a big array of expected outputs in different orders, probably legacy ruby 1.8 stuff when hash where unordered). So I just modified existing specs here to make it simple.

I'll probably make another PR to simplify those specs after I confirm it was only for ruby 1.8.

spec/addressable/template_spec.rb
```diff
@@ -404,7 +404,7 @@
     {
       :var => "value",
       :semi => ";",
-      :year => %w(1965 2000 2012),
+      :year => [1965, 2000, 2012],
```
jarthod, line 407: And this was a good existing example to test integers too :+1:

dentarg (approved):

PR #504, opened 2023-04-06
restore unicode_normalize_kc as a deprecated method

jarthod:
As discussed in https://github.com/sporkmonger/addressable/pull/492#discussion_r1157346135, this is the PR to restore `unicode_normalize_kc` as a deprecated method (in case some people where using it). Example of the produced warning:
```
NOTE: Addressable::IDNA.unicode_normalize_kc is deprecated; use String#unicode_normalize(:nfkc) instead. It will be removed on or after 2023-04.
Addressable::IDNA.unicode_normalize_kc called from benchmark/unicode_normalize.rb:17.
```

### DIFF
--- a/benchmark/unicode_normalize.rb
+++ b/benchmark/unicode_normalize.rb
@@ -2,7 +2,7 @@
 # frozen_string_literal: true.
 
 require "benchmark"
-require "addressable/idna/pure.rb"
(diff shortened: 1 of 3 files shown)


dentarg (approved):

Bizzyfortune (commented):
Ok

666Ty6uK666 (approved):

PR #508, opened 2023-04-27
Fix YAML serialization

casperisfine:
Ref: https://github.com/sporkmonger/addressable/pull/486

Following the introduction of `NONE` flags, `Addressable::URI` instance can no longer be correctly serialized with YAML.

It appear to work, but `NONE` is serialized as `!ruby/object {}`, so when the YAML is parsed back, the attributes are assigned a distinct anonymous object.

FYI: @tenderlove

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -2396,6 +2396,25 @@ module Addressable
       @validation_deferred = false
     end
 
+    def encode_with(coder)
+      instance_variables.each do |ivar|
+        value = instance_variable_get(ivar)
+        if value != NONE
+          key = ivar.to_s.slice(1..-1)
+          coder[key] = value
+        end
+      end
+      nil
+    end
+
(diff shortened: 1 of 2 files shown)


tenderlove:
I'd need to examine the values, but maybe we could use a symbol or number instead of a singleton object instance? Then we don't need to implement the yaml coder stuff

dentarg:
Is #509 needed if this is merged?

casperisfine:
> Is #509 needed if this is merged?

Not strictly, but I think it would be desireable. #509 add backward compatibility, allowing processes running older versions of `assressable` to deserialize YAML generated with newer version, which is important to allow a rollout without downtime.

dentarg:
@tenderlove did you have any other idea or are you okay with this?

tenderlove:
@dentarg I don't have any better ideas, I think we should merge this 😅

PR #509, opened 2023-05-02
Define URI::NONE as a module to avoid serialization issues

casperisfine:
If it's an anonymous objects, when serialized and restored by Marshal or YAML, the restored instance has a diferent anonymous object as ivar values.

The simplest fix for this is to use a named module.

This is related to https://github.com/sporkmonger/addressable/pull/508, but only focus on current version compatibility.

cc @tenderlove

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -2569,7 +2569,7 @@ module Addressable
       @query = nil
     end
 
-    NONE = Object.new.freeze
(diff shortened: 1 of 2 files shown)


dentarg:
@tenderlove does this look good to you?

tenderlove:
@dentarg yep, it looks good to me!

dentarg:
Thanks @tenderlove

PR #510, opened 2023-05-10
Add public_suffix method to allow users to access the PublicSuffix::Domain object easily.

excid3:
This exposes the `PublicSuffix::Domain` object so users can interact with that as needed. Plus it provides a consistent interface for Addressable::URI to interact with.

This is useful for accessing subdomains. PublicSuffix's `#trd` method returns `"www.test"` from `"www.test.example.org"`. 

I'd also like to add `subdomain` & `subdomains` methods to Addressable::URI for direct access to subdomains (which I have a follow up Pull Request for if this is accepted).

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -1198,7 +1198,7 @@ module Addressable
     # @example
     #   Addressable::URI.parse("http://www.example.co.uk").tld # => "co.uk"
     def tld
-      PublicSuffix.parse(self.host, ignore_private: true).tld
+      public_suffix.tld
     end
 
     ##
@@ -1216,7 +1216,16 @@ module Addressable
     # @example
     #   Addressable::URI.parse("http://www.example.co.uk").domain # => "example.co.uk"
     def domain
-      PublicSuffix.domain(self.host, ignore_private: true)
+      public_suffix.domain
+    end
+
+    ##
+    # Returns a PublicSuffix::Domain object for this host.
+    #
+    # @example
+    #   Addressable::URI.parse("http://www.example.org").public_suffix #=> #<PublicSuffix::Domain @sld="example", @tld="org", @trd="www">
+    def public_suffix(**kwargs)
+      PublicSuffix.parse(self.host, **{ignore_private: true}.merge(kwargs))
     end
 
     ##
--- a/spec/addressable/uri_spec.rb
+++ b/spec/addressable/uri_spec.rb
@@ -6799,3 +6799,9 @@ describe Addressable::URI, "when deferring validation" do
     expect(res).to be nil
   end
 end
+
+describe Addressable::URI, "public suffix" do
+  it "should return a PublicSuffix::Domain object" do
+    expect(Addressable::URI.parse("http://example.org").public_suffix).to be_an(PublicSuffix::Domain)
+  end
+end


dentarg:
> This exposes the PublicSuffix::Domain object 

a bit hesitant about this, sure this object looks simple enough now, but what if it changes down the road? There could be unexpected breaking changes for users of addressable then

excid3:
Addressable already relies on it, so it would need to be updated for breaking changes as well. It also seems pretty unlikely to change since it only handles parsing domains.

I've been working with a lot of apps that allow custom domains & subdomains and it would be really helpful for Addressable to support parsing those out. I was surprised that wasn't a feature considering it was already using PublicSuffix.

I can merge the `subdomain` & `subdomains` feature into this branch if you'd prefer to consider it all together.

excid3:
Decided to merge those into this PR to show the complete feature.

dentarg:
> Addressable already relies on it, so it would need to be updated for breaking changes as well. 

Yes, but we own the API, and I think it is a difference in interacting with PublicSuffix to get a basic value (e.g. String) then returning a full PublicSuffix object.

> It also seems pretty unlikely to change since it only handles parsing domains.

Yeah, `PublicSuffix::Domain` probably wont change but let's not take any chances.

excid3:
What do you think works best? Should we keep the `subdomain` and `subdomains` methods and make `public_suffix` private?

dentarg:
I'm not sold on these additions. I'm feel there's ambiguity around the terms "subdomain" and "subdomains"... this looks a bit different from what publicsuffix-ruby does? 

```ruby
irb(main):006:0> PublicSuffix.parse("www.google.com", ignore_private: true).subdomain
=> "www.google.com"
```

lib/addressable/uri.rb
```diff
@@ -1216,7 +1216,34 @@ def tld=(new_tld)
...
+    def subdomains
+      public_suffix.trd.to_s.split(".")
+    end
+
+    ##
+    # Returns a PublicSuffix::Domain object for this host.
+    #
+    # @example
+    #   Addressable::URI.parse("http://www.example.org").public_suffix #=> #<PublicSuffix::Domain @sld="example", @tld="org", @trd="www">
+    def public_suffix(**kwargs)
```
dentarg, line 1245: I'm thinking it is better that users interact directly with `PublicSuffix` if they have the need to customise the options.

excid3:
> I'm not sold on these additions. I'm feel there's ambiguity around the terms "subdomain" and "subdomains"... this looks a bit different from what publicsuffix-ruby does?

That was on purpose. I wanted a way to access the subdomains (without the primary domain). 

The `subdomain` method from PublicSuffix doesn't really add any value since it's basically the same as `host` in Addressable.

lib/addressable/uri.rb
dentarg: Let's make this method private, use memoization (but cleared if `#host=` is used), and remove the arguments

lib/addressable/uri.rb
```diff
@@ -1216,7 +1216,34 @@ def tld=(new_tld)
...
+      public_suffix.domain
+    end
+
+    ##
+    # Returns the first subdomain for this host.
+    #
+    # @example
+    #   Addressable::URI.parse("http://www.test.example.co.uk").domain # => "www"
+    def subdomain
+      subdomains.first
```
dentarg, line 1228: Still think there's ambiguity around what's a [subdomain](https://en.wikipedia.org/wiki/Subdomain) and what a method like this is expected to return. 

`www` is a subdomain of `test.example.co.uk`
`test` is a subdomain of `example.co.uk`
`example` is a subdomain of `co.uk`

This method would return `nil` when parsing `http://example.co.uk`

dentarg: I'm not familiar with that in Rails, can you link the source for method? When is it typically used?

I'm not sure what we want here, if the method should exist?

excid3: Multi-tenant applications or anything that uses subdomains (think Shopify) use this. I've always been surprised there's no way to get the subdomain(s) from URIs in Ruby. Rails will do it for the current request, but that's it and it's not Public Suffix aware. You can only specify a single TLD for your entire Rails app, so you need to reach for something else like addressable + public_suffix.

https://api.rubyonrails.org/classes/ActionDispatch/Http/URL.html#method-i-subdomain

lib/addressable/uri.rb
excid3: Done!

lib/addressable/uri.rb
excid3: Any thoughts on this?

dentarg: Yeah... _if_ we should add this method, is it not better if it matches the `subdomain` on ActionDispatch? From the URL above
> Returns all the subdomains as a string, so "dev.www" would be returned for “dev.www.rubyonrails.org”. 

For `http://www.test.example.co.uk` that would be `www.test`, not just `www`?

lib/addressable/uri.rb
```diff
@@ -1216,7 +1217,25 @@ def tld=(new_tld)
...
     def domain
-      PublicSuffix.domain(self.host, ignore_private: true)
+      public_suffix.domain
+    end
+
+    ##
+    # Returns the first subdomain for this host.
+    #
+    # @example
+    #   Addressable::URI.parse("http://www.test.example.co.uk").domain # => "www"
```
dentarg, line 1227:
```suggestion
    #   Addressable::URI.parse("http://www.test.example.co.uk").subdomain # => "www"
```

lib/addressable/uri.rb
```diff
@@ -1216,7 +1217,25 @@ def tld=(new_tld)
...
+    def subdomain
+      subdomains.first
+    end
+
+    ##
+    # Returns the subdomains for this host.
+    #
+    # @example
+    #   Addressable::URI.parse("http://www.test.example.co.uk").domain # => ["www", "test"]
```
dentarg, line 1236:
```suggestion
    #   Addressable::URI.parse("http://www.test.example.co.uk").subdomains # => ["www", "test"]
```

spec/addressable/uri_spec.rb
```diff
@@ -6799,3 +6799,23 @@ def to_str
...
+
+describe Addressable::URI, "subdomains" do
+  it "should return the subdomains" do
+    expect(Addressable::URI.parse("http://www.test.example.org").subdomains).to eq(["www", "test"])
+  end
+
+  it "should return an empty array if no subdomain" do
+    expect(Addressable::URI.parse("http://example.co.uk").subdomains).to eq([])
+  end
+end
```
dentarg, line 6821: Please keep the [terminating newline](https://stackoverflow.com/a/729795/525616)

Screenshot 2023-06-03 at 22 18 31

lib/addressable/uri.rb
dentarg: What about not adding the `#subdomain` method? Instead only `#subdomains`. It is easy for the user to do `subdomains.first` or `subdomains.first(2)` etc. depending on what they want.

dentarg:
@sporkmonger any thoughts on this?

PR #512, opened 2023-06-07
Template: Don't assume mapping is enumerable

Hbutlercapone:
this allows template expansions with lambdas or anything else that responds to `[]`


therabidbanana (approved):

therabidbanana:
https://github.com/sporkmonger/addressable/pull/133#issuecomment-25284634

This point was made with an earlier request for this type of feature - I think we probably want to add some additional checks to make sure the API is solid and add a bit more to the docs?

therabidbanana:
An additional commit that covers the requested doc updates, argument errors, and adds more tests for edge cases from this particular update (partial_expand doesn't make sense with a lambda). 

I can set up a new PR with the changes or you can cherry pick my commit into your PR @Hbutlercapone and clean them up some more if you'd like - I don't know how we'd want partial_expand to work here but there's a few ways we could approach it aside from "throw error"

 https://github.com/sporkmonger/addressable/compare/main...therabidbanana:addressable:pr/133

lib/addressable/template.rb
```diff
@@ -922,6 +922,7 @@ def normalize_value(value)
     # @return [Hash]
     #   A hash with stringified keys
     def normalize_keys(mapping)
+      return mapping unless mapping.respond_to?(:inject)
```
therabidbanana, line 925: As I was adding more tests/docs for this I realized it's incomplete for `#partial_expand` which requires something responding to `#key?` because nil is a valid expansion value (which becomes an empty string)

therabidbanana (changes requested):
Added a commit with docs and some tests to cover the edge case with partial_expand - we should include that before merging

Hbutlercapone:
I will hop on after work tonight and take a look!

Hbutlercapone (commented):
.

Issue #514, opened 2023-07-18
Unsafe concurrent Hash access


eregon:
It happens for SEQUENCE_UPCASED_PERCENT_ENCODING_TABLE and SEQUENCE_ENCODING_TABLE
See https://github.com/oracle/truffleruby/issues/3166#issuecomment-1640352210

This happens on TruffleRuby and possibly on JRuby.
On CRuby the error will not happen due to the GIL.
On CRuby it may still cause duplicate computation and assignment for the same key.

A good solution would be to use a Concurrent::Map from `concurrent-ruby` instead of raw `Hash` there.
And the same for SEQUENCE_ENCODING_TABLE just above.
(as a note, this is the correct pattern: https://github.com/ruby-concurrency/concurrent-ruby/issues/970)
Then it's fast on CRuby and is thread-safe and scales on no-GVL Rubies.

Of course this would add a dependency on  `concurrent-ruby` for this gem.

eregon:
An alternative would be to use a Mutex around all accesses to these global Hash'es.
But that would be slower on CRuby due to the extra Mutex on read, and not scale on no-GVL Rubies.
So performance-wise it's strictly worse than concurrent-ruby for all Rubies.

dentarg:
> It happens for SEQUENCE_UPCASED_PERCENT_ENCODING_TABLE and SEQUENCE_ENCODING_TABLE

Because I looked it up I'll note that those were introduced with https://github.com/sporkmonger/addressable/commit/880b34c6518279cf9450e4af3818e51531db5a69

dentarg:
> A good solution would be to use a Concurrent::Map from `concurrent-ruby` instead of raw `Hash` there

Would that help even before https://github.com/ruby-concurrency/concurrent-ruby/issues/970 is addressed?

eregon:
Yes, the only thing I meant by that link is one should use that `compute_if_absent` pattern to avoid assigning a given key twice.

dentarg:
Right, so the way to address this (with `concurrent-ruby`) is using `Concurrent::Map` and `compute_if_absent` [like this](https://github.com/rails/rails/pull/46536/files)?

eregon:
Yes

eregon:
More comments by @byroot at https://github.com/oracle/truffleruby/issues/3166#issuecomment-1641498276

eregon:
@byroot Any chance you could take a look at this? :)

It'd be good to figure out what's the desired solution for `addressable`.

[...]

PR #515, opened 2023-07-31
Fix thread safety issue with encoding tables

casperisfine:
Fix: https://github.com/sporkmonger/addressable/issues/514
Fix: https://github.com/oracle/truffleruby/issues/3166

These hashes lazily memoize percent encoded characters, this is an issue on GVL-less Ruby implementations as it can cause concurrent access.

But these are actually quite wasteful as the key is always a single byte so rather than use string keys are lazily memoize these, we can precompute two static arrays of 255 elements once and for all.

@eregon @dentarg 

NB: I'm not a big addressable user, so I mostly relied on the test suite, hopefully my assertions are correct.


lib/addressable/uri.rb
```diff
@@ -560,10 +557,10 @@ def self.normalize_component(component, character_class=
         leave_re = if leave_encoded.length > 0
           character_class = "#{character_class}%" unless character_class.include?('%')
 
-          "|%(?!#{leave_encoded.chars.flat_map do |char|
-            seq = SEQUENCE_ENCODING_TABLE[char]
-            [seq.upcase, seq.downcase]
-          end.join('|')})"
+          bytes = leave_encoded.bytes
+          leave_encoded_downcase = bytes.map { |b| SEQUENCE_ENCODING_TABLE[b] }.join('|')
+          leave_encoded_upcase = bytes.map { |b| SEQUENCE_UPCASED_PERCENT_ENCODING_TABLE[b] }.join('|')
```
eregon, line 562: `SEQUENCE_UPCASED_PERCENT_ENCODING_TABLE` has extra `%`:
```
SEQUENCE_ENCODING_TABLE[
> format("%02x", 45)
=> "2d"
SEQUENCE_UPCASED_PERCENT_ENCODING_TABLE
> format("%%%02X", 45)
=> "%2D"
```
So it should be without the `%` because that's already present outside (`|%`)

eregon (commented):
This looks much simpler, thank you!

lib/addressable/uri.rb
casperisfine: Hum, I think you are right, but it's weird because I didn't change that 🤔

casperisfine: Ah nevermind, I get it now.

casperisfine: Should be good now.

dentarg:
What do we think about the ruby-head failure? [`double free or corruption (out)`](https://github.com/sporkmonger/addressable/actions/runs/5713300669/job/15478461111?pr=515#step:7:11) (maybe it is unrelated to this change, [we'll see](https://github.com/dentarg/addressable/actions/runs/5716126315/job/15486934643))

casperisfine:
> maybe it is unrelated to this change

It is. `Addressable` being pure Ruby, it can in no way being responsible for a VM crash of this sort.

PR #522, opened 2023-09-13
Link directly to versioned changelog from gemspec

stoivo:
I like it a lot when the changelog url on rubygems link directly to the versions I updated to

### DIFF
--- a/tasks/gem.rake
+++ b/tasks/gem.rake
@@ -31,7 +31,7 @@ namespace :gem do
     s.homepage = "https://github.com/sporkmonger/addressable"
     s.license = "Apache-2.0"
     s.metadata = {
-      "changelog_uri" => "https://github.com/sporkmonger/addressable/blob/main/CHANGELOG.md"
+      "changelog_uri" => "https://github.com/sporkmonger/addressable/blob/addressable-#{PKG_VERSION}/CHANGELOG.md"
     }
   end
 
--- a/tasks/gem.rake
+++ b/tasks/gem.rake
@@ -31,12 +31,7 @@ namespace :gem do
     s.homepage = "https://github.com/sporkmonger/addressable"
     s.license = "Apache-2.0"
     s.metadata = {
-      # We want url like below. If we have version 3.11.0 and 3.1.10. When only
-      # one exist it would have id addressable-3110, when the secund tag also
-      # exist it would create addressable-3110 and addressable-3110_1. In this
-      # case we would might get incorrect anchors
-      # https://github.com/sporkmonger/addressable/blob/main/CHANGELOG.md#addressable-285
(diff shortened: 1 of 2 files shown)


dentarg:
I'm not sure about this change, the changelog can be updated after release too. For example, if we forgot something.

stoivo:
I didn't think, would be nice handle it somehow. I have a new idea. Since the title don't change we can do `https://github.com/sporkmonger/addressable/blob/main/CHANGELOG.md#addressable-284`. If we we this with no other changes we will have issued with referencing the correct version. Update changelog.log with  have titles like `#Addressable 2.1.10 <a name="addressable-v2.1.10">` then we can link to it with `https://github.com/sporkmonger/addressable/blob/main/CHANGELOG.md#user-content-addressable-v2.1.10">`.

Should I update the PR to this so you can see it for yourself?

dentarg:
Sure, that's a good idea. Not sure I follow the last part though, does the actual CHANGELOG file need updating too? It would be best if it could be kept as-is and we make the changelog link be on the already auto-generated format `https://github.com/sporkmonger/addressable/blob/main/CHANGELOG.md#addressable-284`

sporkmonger:
I think it should be linked via the release tag, right?

stoivo:
For all other project where I have seen changelog links to specific version, it has linked to the tag but as dentarg mentioned. We might want to add something to the changelog after the tag was made

stoivo:
So I have make some improvements. Which one do you prefer, should I drop the last commit?

dentarg:
Sure, this looks good to me

I will just point out that it seems to work without `user-content-` part too: https://github.com/stoivo/addressable/blob/patch-1/CHANGELOG.md#v2.5.0

tasks/gem.rake
```diff
@@ -31,7 +31,9 @@ namespace :gem do
     s.homepage = "https://github.com/sporkmonger/addressable"
     s.license = "Apache-2.0"
     s.metadata = {
-      "changelog_uri" => "https://github.com/sporkmonger/addressable/blob/main/CHANGELOG.md"
+      # We want url like below.
+      # https://github.com/sporkmonger/addressable/blob/main/CHANGELOG.md#user-content-v2.8.5"
```
dentarg, line 35: I don't think we need these comments

tasks/gem.rake
```diff
@@ -31,7 +31,9 @@ namespace :gem do
...
+      "changelog_uri" => "https://github.com/sporkmonger/addressable/blob/main/CHANGELOG.md#user-content-v#{PKG_VERSION}"
```
dentarg, line 36: I think this is cleaner as it works too

```suggestion
      "changelog_uri" => "https://github.com/sporkmonger/addressable/blob/main/CHANGELOG.md#v#{PKG_VERSION}"
```

stoivo:
it is updated

PR #533, opened 2024-05-21
Add bigdecimal to Gemfile

m-nakamura145:
From Ruby 3.4 onwards, gems that are expected to become bundled gems will raise a LoadError if they are required without being present in the Gemfile or gemspec. Therefore, I have added the target gem to the Gemfile.

```sh
/addressable/spec/addressable/template_spec.rb:20: warning: bigdecimal was loaded from the standard library, but is not part of the default gems since Ruby 3.4.0. Add bigdecimal to your Gemfile or gemspec.

An error occurred while loading ./spec/addressable/template_spec.rb. - Did you mean?
                    rspec ./spec/addressable/idna_spec.rb
                    rspec ./spec/addressable/uri_spec.rb
                    rspec ./spec/addressable/security_spec.rb

Failure/Error: require "bigdecimal"

LoadError:
  cannot load such file -- bigdecimal
# ./spec/addressable/template_spec.rb:20:in '<top (required)>'
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ source 'https://rubygems.org'
 gemspec
 
 group :test do
+  gem 'bigdecimal' if RUBY_VERSION > '2.4'
   gem 'rspec', '~> 3.8'
   gem 'rspec-its', '~> 1.3'
 end


.github/workflows/test.yml
```diff
@@ -89,6 +89,7 @@ jobs:
           - 3.1
           - 3.2
           - 3.3
+          - 3.4.0-preview1
```
dentarg, line 92: We don't need to add this preview version, we have a job for `head`, which now is `ruby 3.4.0dev` (https://github.com/sporkmonger/addressable/actions/runs/9173838413/job/25223381402#step:5:25)

Last time that job ran was 4 months ago, so I suspect the `bigdecimal` change had not happen yet as it passed back then (but the logs are now gone)

Gemfile
```diff
@@ -5,6 +5,7 @@ source 'https://rubygems.org'
 gemspec
 
 group :test do
+  gem 'bigdecimal'
```
dentarg, line 8: You need to make this work for the jobs that are failing

dentarg (approved):

Issue #527, opened 2023-12-09
Improve release flow


dentarg:
It is too much cruft, it is time to make it easier and more automated

* Stop generating the gemspec, update it by hand instead
* Remove all the old rake tasks no longer needed
* Probably use https://github.com/svenfuchs/gem-release
* Push new versions to RubyGems.org from GitHub Actions (https://github.com/rubygems/rubygems.org/issues/4126)
  * already possible to use, I've tested at https://github.com/spinels/spinels/blob/main/.github/workflows/release.yml

dentarg:
Examples

* All PRs wanting to change the gemspec (misses there's a rake task to update)

* https://github.com/sporkmonger/addressable/pull/526

* ```shell
  $ b e rake --trace clobber
  ** Invoke clobber (first_time)
  rake aborted!
  Don't know how to build task 'spec:clobber_simplecov' (See the list of available tasks with `rake --tasks`)
  Did you mean?  spec:simplecov
  ...
  ```

dentarg:
https://blog.rubygems.org/2023/12/14/trusted-publishing.html

sporkmonger:
The new trusted publishing flow is the way to go I think.

sporkmonger:
Might be worth exploring SLSA provenance too if we're making release changes.

dentarg:
> Might be worth exploring SLSA provenance too if we're making release changes.

It will come: https://github.com/rubygems/rubygems.org/issues/4286#issuecomment-1854941267

dentarg:
I tested with "Trusted Publishers" with a dummy gem over at https://github.com/spinels/slenips

dentarg:
It worked 🎉  https://github.com/sporkmonger/addressable/actions/runs/9612404033/job/26512966861#step:8:13

> Successfully registered gem: addressable (2.8.7)

PR #536, opened 2024-06-21
Add Trusted Publishing workflow

dentarg:
Part of #527


dentarg:
Note to self: don't use `[ci skip]` in commit message, that will cause status to never be reported to the pull request, blocking merge (only possible when CI is passing)

PR #545, opened 2024-08-13
fix(uri): query_value guetter - return all elements in array not just last value

SebastienLAURET:
Hi, 

I have a problem when i want to use `query_values`.
When a query params is an array this lib return only the last element 

```ruby
uri = Addressable::URI.parse("http://my_url.com?param=a&param=b")
uri.query_values 
=> {
  param: "b"
}
```

but I think it better if it return all values in array like that :

```ruby
uri = Addressable::URI.parse("http://my_url.com?param=a&param=b")
uri.query_values 
=> {
  param: ["a", "b"]
}
```

So i propose an implementation for this problem

Thank's

### DIFF
--- a/lib/addressable/uri.rb
+++ b/lib/addressable/uri.rb
@@ -1693,7 +1693,7 @@ module Addressable
           pair[1] = URI.unencode_component(value)
         end
         if return_type == Hash
-          accu[pair[0]] = pair[1]
+          accu[pair[0]] = accu.key?(pair[0]) ? [accu[pair[0]], pair[1]].flatten : pair[1]
         else
           accu << pair
         end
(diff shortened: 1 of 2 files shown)


dentarg:
From https://github.com/sporkmonger/addressable/issues/282#issuecomment-369704352, you can do this:

```ruby
$ irb -raddressable/uri
irb(main):001:0> uri = Addressable::URI.parse("http://my_url.com?param=a&param=b")
irb(main):002:0> uri.query_values
=> {"param"=>"b"}
irb(main):003:0> uri.query_values(Array)
=> [["param", "a"], ["param", "b"]]
```

